Issue #2, opened 2015-03-11
Add CocoaPods support


Katafalkas:
:+1:

nuclearace:
The problem is, I have no idea how to use CocoaPods. I understand the benefits, I've just found it easier for Swift based things to just use copy/paste.

nuclearace:
Added support

Issue #3, opened 2015-03-11
Maybe use Starscream instead ?


Katafalkas:
https://github.com/daltoniam/starscream Is fully swift.

nuclearace:
The reason I didn't use starscream goes back to #2. And from the fact that I've had trouble trying to use full Swift frameworks. Although looking at the API, it is nice that it separates between Strings and Data. Which would simplify some of the client.

nuclearace:
Another reason I didn't use starscream is that using Swift frameworks automatically means you're limited to iOS 8+.

nuclearace:
Actually, it's really easy to just include starscream. So I might do that

PR #4, opened 2015-03-11
Add CocoaPods podspec

banjun:
refs #2 

I have added a podspec and a example project, although there are some difficulty because SocketRocket's current commits are not published to the master CocoaPods/Specs repo...
please edit the URLs, authors and tags in the podspec file and LICENSE file (stub).


nuclearace:
I'm actually in the process of moving to starscream per #3.

nuclearace:
Also, using CocoaPods would mean iOS 8+ only, and I'm not sure if I want to do that right now.

Katafalkas:
How come ? Do you have a link to that information ? Cant seem to find it.

nuclearace:
http://blog.cocoapods.org/CocoaPods-0.36/

nuclearace:
Basically you can't have dynamic libraries in iOS 7. And you can't build static Swift libraries.

banjun:
Yes, CocoaPods does not support Swift pods on iOS 7.
While theoretically we can make a swift library hybrid for both as a submodule and as a pod, I understand that it may need some more work to maintain.

Anyway, I rebased on Starscream commits. this should work with both CocoaPods and submodule.
https://github.com/banjun/socket.io-client-swift/tree/cocoapods-starscream

I think this PR 4 may be closed because this conflicts with the current master.

nuclearace:
@banjun if you want to open another that will merge. I don't have time to manually sort out conflicts.

nuclearace:
Also, you don't really need to include an example. One is linked in the README

banjun:
I'll rebase and open another one without the example. Thanks!

Issue #14, opened 2015-03-16
Swift 1.2 PodSpec branch.


Katafalkas:
After installing `socket.io-client-swift` with Podfile - the installed version of this lib was from master branch, not from 1.2 branch. How do we change it ?

bjonica:
pod 'Socket.IO-Client-Swift', :git => 'https://github.com/socketio/socket.io-client-swift.git', :branch => '1.2'

PR #15, opened 2015-03-16
changing the branch for pod installation.

Katafalkas:
Did not test it, but according to the `http://guides.cocoapods.org/using/the-podfile.html` - this should work.
Fixes socketio/socket.io-client-swift#14


Katafalkas:
Fixes socketio/socket.io-client-swift#14

nuclearace:
I wouldn't treat the 1.2 branch as stable enough to warrant defaulting to it. When 1.2 is released I'll merge the 1.2 branch into master

Katafalkas:
as I understand currently pods are enabled only for 1.2 branch, right ? if so - whats the harm of changing podspces just for 1.2 branch ? Or what am I missing ?

nuclearace:
Currently CocoaPods gets from the master branch

PR #18, opened 2015-03-18
updated parsing to better handle acks

hanishabsigh:


nuclearace:
This seems kind of redundant. Is something breaking?

hanishabsigh:
Yes, I wasn't receiving an ack for some of my messages because the first regex was outputting some sort of groups. Plus, this seems more efficient.

nuclearace:
I think I'm going to redo parsing

nuclearace:
Since it's basically the same as it's been for months

nuclearace:
Master should handle parsing better

PR #28, opened 2015-03-26
Carthage Support

justinmakaila:
- Added support for building the framework using Carthage. 

- Renamed the SwiftIO directory to SocketIO for consistency. 

- Includes OS X support, I didn't dive into the code to see if there's anything platform dependent, but skimming through it looked OK. 

- Updated the README for carthage installation directions.

- The pod spec should probably be reviewed, as I'm not a user of cocoapods, but it should be correct. 

This does **NOT** include any of the updates for Swift 1.2, and should be merged in before any of those changes are made.


justinmakaila:
Also added a new v1.4.2 tag

nuclearace:
Does Carthage require a whole Xcode thing. Because I explicitly made it so I wouldn't be committing one.  Also, it's not really adding OS X support. I actually don't test at all on iOS really. All testing I do is from OS X.

nuclearace:
The reason the project has CocoaPods support in the first place is someone added it. Frankly, I hate all the dependency managers for Xcode, and I prefer to use git for dependency tracking.

nuclearace:
I do agree that the SwiftIO folder should be renamed. It's actually something I've been meaning to do, just been too lazy to do it.

justinmakaila:
Carthage requires a Run Script build phase to strip x86 symbols when submitting to the App Store, but needs nothing extra committed to the repo. You can read more [here](https://github.com/Carthage/Carthage).

By OS X support, I meant there's now an OS X target, which Carthage uses to build, and technically, a library does not support Carthage unless it builds for iOS and OSX.

nuclearace:
>Carthage will only build Xcode schemes that are shared from your .xcodeproj

I really don't want to include a whole .xcodeproj. I hate that kind of thing in a github repo. I'd rather stick to CocoaPods and/or git.

tl;dr: I hate Xcode dependency managers.

justinmakaila:
I did the work for you. I respect if you don’t want to pull it, as I’ll just use my own fork. But the option is there.


—Justin Makaila

nuclearace:
I'd add it, the problem is I don't really know enough about it so that if someone says it's broken, how I'd go about fixing it.

justinmakaila:
All it is is an added Xcode project which wraps the files so they can be built as a `.framework`. 


—Justin Makaila

nuclearace:
If you can merge the latest master, I'll merge your branch.

justinmakaila:
Word, I’ll get that in a bit and comment back here when it’s done.


—Justin Makaila

nuclearace:
Does carthage support 1.2

justinmakaila:
Carthage is nothing but a fancy CLI for wrapping the `xcodebuild` tools to checkout and build all of your dependencies.

—Justin Makaila

justinmakaila:
Merged in master and began updating some documentation and Swift style stuff. I'll continue adding documentation and style guide conformance over time to my fork, but this is ready to go to get carthage in.

Carthage is supported for releases >= v2.0.2

nuclearace:
Cool, I'll look into this later today. Just some style things that trigger me, but I can change back.

While the style Apple uses is

```swift
var variable: Type
```

I generally don't like this. Because : is used a bunch, I prefer to only use a space after : when it's used in a function call, not type declaration.

justinmakaila:
I find that extremely difficult to read. Especially with the lack of whitespace.


—Justin Makaila

nuclearace:
Hm, I'm going to have to test around with this a bit. I'm still unsure if Carthage support is something I want to incorporate.

justinmakaila:
I recommend setting up a new project, going through CocoaPods integration, and then going through Carthage integration.




The benefits, IMO, are obvious. Everything will live inside of it’s own namespace and framework, so you can `import SocketIO` rather than having everything live in the global namespace. Even if users don’t use carthage, they can still open up the `.xcodeproj`, build the project, and then drop that `.framework` into their project.




I’m not a fan of dropping tons of `.swift` files into my project.




I also believe, being the main SocketIO repository, you’d be doing a disservice to the entire iOS development community by rejecting a pull request which adds support for one of the main dependency managers based on personal bias. 


—Justin Makaila

Xaekai:
You sure do love your whitespace.

justinmakaila:
Yes, yes I do :+1:. However, if we’re talking about that last comment, that was my email client.


—Justin Makaila

nuclearace:
The problem is once this is merged, I don't really want to keep up with another dependency manager. And I don't even know how I would keep the carthage one updated

nuclearace:
Since it makes use of Xcode schemas.

justinmakaila:
You don't have to do anything. If you do, personally mention be, because at this rate, I'll be maintaining my own fork.



—Justin Makaila

justinmakaila:
Here's another link to the Carthage repo: http://github.com/Carthage/Carthage. You can find everything you want to know there.



—Justin Makaila

nuclearace:
I'm in the process of merging this in another branch

nuclearace:
https://github.com/socketio/socket.io-client-swift/tree/carthage

nuclearace:
I tested to see it would build frameworks. Didn't test if they actually work. I've had difficulties with dynamic frameworks in the past

justinmakaila:
They work, otherwise I wouldn’t have made the pull request. I’m currently using them, built by Carthage, in a production app.


—Justin Makaila

nuclearace:
I mean the ones in the merged branch, I wasn't sure if I messed something up.

nuclearace:
I want to merge the changes to the source files separately from the carthage support. So I'm merging from another branch. https://github.com/socketio/socket.io-client-swift/pull/50 Comment on Carthage here

PR #53, opened 2015-04-13
Fixed the use of the ?? operator because Swift is broken

CallumOz:
I fixed a strange bug where in `handleAck`, the data would be inclosed in an unnecessary Array because of a bug in swift/Xcode/clang.

When you use the `??` operator in the call to function, it doesn't seem to detect the lack of parenthesis to inform of the desired priority.

This screenshot should help explain:

screen shot 2015-04-13 at 15 40 43

As you can see, line 11 isn't flagged as an error.
Line 13 is my fix by adding parenthesis.
And line 15 generates an error.


nuclearace:
Damn operator precedence.

PR #59, opened 2015-04-21
[FIX] memory leak with the takeUnretainedValue()

yannickl:
Hi,

I know the file came from the `Starscream` project (and I'll push the same pull request) but I found a memory leak inside the `validateResponse` method with `takeUnretainedValue()` calls.

Here the `CFHTTPMessageCreateEmpty` and the `CFHTTPMessageCopyAllHeaderFields` needs to be manage by swift so we need to call `takeRetainedValue()` instead.

Best


nuclearace:
Interesting, I haven't really done any leakage checking in a while.

Issue #111, opened 2015-07-06
Emit Dictionary in Objective-C


TalhaAkkas:
In objective-c emitObj (active pod) method gives warning.
I patched it for current repo.  
![screen shot 2015-07-06 at 09 37 58](https://cloud.githubusercontent.com/assets/632596/8516885/ba64d71e-23c2-11e5-91b9-8e44dc67b209.png)

nuclearace:
This isn't really a problem. The reason it takes an array is because socket.io lets you send multiple items per message. With Swift it's easy to get around using variadic parameters. I'm not aware of an easier Objective-C solution than to use the equivalent array of AnyObject.

```swift
// Swift
public func emit(event:String, _ items:AnyObject...)

// Objective-C
public func emit(event:String, withItems items:[AnyObject])
```

PR #112, opened 2015-07-06
emit dictionary in Objective-C

TalhaAkkas:
My patch for issue #111

### DIFF
--- a/SocketIOClientSwift/SocketIOClient.swift
+++ b/SocketIOClientSwift/SocketIOClient.swift
@@ -292,6 +292,18 @@ public final class SocketIOClient: NSObject, SocketEngineClient, SocketLogClient
             self?._emit(event, items)
         }
     }
+    /**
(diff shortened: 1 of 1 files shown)


nuclearace:
It really isn't an issue. The reason it takes an nsarray is because with socket.io you can send multiple items with one message.

TalhaAkkas:
of course not but my api for example in server side expects given item to be a object not array so i need to send dictionary. so some body can be need it too

nuclearace:
It won't be sent as an array, the client will go through each item in the array and parse it. The Swift and Objective-C APIs are functionally the same, Swift just adds some syntactic sugar that will turn

```swift
socket.emit("event", 2, 3, ["test": test"])
```

into

```swift
socket.emit("event", [2, 3, ["test": test"]])
```

But this doesn't exist in Objective-C (that I'm aware of).

PR #117, opened 2015-07-12
Added project with iOS and Mac targets => Carthage support

NachoSoto:
This enables users to install the library using [Carthage](https://github.com/Carthage/Carthage).

Additionally, the repo would now include the necessary configuration settings: things like base SDK, deployment target, optimization levels, additional build flags, etc.

Please let me know if there's any setting you'd like to be different.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,21 @@ Icon
 Network Trash Folder
 Temporary Items
 .apdisk
+
+# Xcode
+build/
+*.pbxuser
+!default.pbxuser
+*.mode1v3
+!default.mode1v3
+*.mode2v3
+!default.mode2v3
+*.perspectivev3
+!default.perspectivev3
+xcuserdata
+*.xccheckout
+*.moved-aside
+DerivedData
+*.hmap
+*.ipa
+*.xcuserstate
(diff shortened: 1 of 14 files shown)


nuclearace:
I really have mixed feelings about Carthage. I prefer to not include Xcode stuff in the repo. That's why I had a separate branch for Carthage.

NachoSoto:
Carthage feelings aside, it's worth noting that source-code is not everything in a library. Like I mentioned, things like base SDK, deployment target, optimization levels, additional build flags, etc. are just important, and an Xcode project is the canonical way in Cocoa to represent and persist these additional details.

If you really disagree about this, it would be great at least to provide built frameworks with the project releases, here in Github, for iOS, Mac, and watchOS.

I predic, however, that this would add more work for you.

Thank you for your time.

nuclearace:
I'll look into it again tomorrow.

nuclearace:
@NachoSoto Have you checked that this doesn't break CocoaPods? If it doesn't break CocoaPods I'll merge it in. The only thing I'm going to have to work out is how I should develop it (properly) without breaking the Carthage support. (I'm not a Cocoa/iOS/OS X dev, I just know Swift fairly well). And the last time I tried working with Carthage the frameworks it output always seemed to be broken for me (I was probably doing something wrong)

nuclearace:
Doesn't seem to break it.

NachoSoto:
@nuclearace glad to hear! Thanks for the merge :)

> The only thing I'm going to have to work out is how I should develop it (properly) without breaking the Carthage support. (I'm not a Cocoa/iOS/OS X dev, I just know Swift fairly well). And the last time I tried working with Carthage, the frameworks it output always seemed to be broken for me (I was probably doing something wrong)

The good thing about Carthage is that, as long as the project builds, you know that Carthage supports is not broken.

In any case, I'll be glad to help if anything comes up. I'm going to be using this library pretty often now so I look forward to helping more!

PR #127, opened 2015-07-24
Tests for socket.io client

lightsprint09:
I really like your project and want to try to contribute to it. But before contributing with confidence I think it is a good start to write some tests.

I started writing tests. I wrote a little [test server] (https://github.com/lightsprint09/socket.io-client-swift-test-server)
I plan to run the server on a free tier on heroku and use Travis when it supports Xcode 7.x. The setup brings continuous integration for free.

By cloning the server repository everyone can run it locally.

I hope you like the idea of building up a set of tests to verify the socket.io client.


nuclearace:
Some tests would be probably be good. Right now this is what I'm using for testing. https://gist.github.com/nuclearace/12c52b9fe910a9402fcc It's pretty crude at the moment. But it gives an idea of how I was testing it.

lightsprint09:
Ok. If it is ok I´ll give it a try to convert in into real test cases.

lightsprint09:
I added the basic tests for emiting many types of data. It would be great to get some feedback

nuclearace:
You can probably remove the cytu.be things in the tests. I only use that to make sure I haven't completely broken everything

lightsprint09:
I am going to refactor the tests and reopen the pr

lightsprint09:
I pushed a new version.
It contains test cases for various datatypes. All of them are tested with emit and emitWithAck.
Some tests do not pass. There are some problems sending Buffer with an acknowledgment and some tests only pass sometimes or when executed exclusively.

I also pushed a new version of the server tests.

I hope we can fix the remaining tests and merge it into master/swift2

nuclearace:
Can we include the server as well?

lightsprint09:
In the Swift repository? It is possible but does it make sense

nuclearace:
How else would anyone run the tests?

lightsprint09:
Cloning the test server separately. But this is not the best way.
Is there a possibility in git to link to the other repo?

nuclearace:
Is there an issue with just have a TestServer directory in this repo?

lightsprint09:
No. I don't think so.
I copy everything and commit it

nuclearace:
I'm getting some errors while testing. They look like it could actually be Swift not compiling things correctly. Does this happen to you?

nuclearace:
Like this


screen shot 2015-07-28 at 11 16 36 am

nuclearace:
Actually nvm. I think it's something with the server.

lightsprint09:
This is actually some kind of bug. I noticed that sending Buffers back cause los of some data... I don't know why

lightsprint09:
At some point there is one data item missing coming from the server. It is always missing when there is a Buffer is involved

nuclearace:
This is really weird. And it's not really easy to pick apart what's going on, at least for me. I'm not used to doing iOS/OS X unit testing.

lightsprint09:
My abstraction is not as is good as it should.

I could remove the Buffers from the server and it will work. But the bug with the buffer will stay...

nuclearace:
What's weird is when I try and test the buffer myself using another server it seems to work.

nuclearace:
But I'm looking at your server in debug mode it it's sending correctly. So it must be something weird happening on the client.

lightsprint09:
Did you test it with emit or ack? I think only ack is the problem. Emit works finde

nuclearace:
I bet I know what's happening. It's a bug in the client.

lightsprint09:
It could be. But it could be a bug in the bad test implementation.

nuclearace:
Maybe, TBH I'm not really sure what's going on.

nuclearace:
I'll have a closer look tonight.

lightsprint09:
This is a good idea. Take some time...

nuclearace:
What's weird for the tests that fail is that it looks like it's not even really emitting those events. Like it looks like it's handling the `connect` event when it really should be handling `emitJSON`

nuclearace:
I feel like weird things are happening with the tests.

nuclearace:
Specifically with the ack tests.

lightsprint09:
I will take a look at all the tests again. Maybe I can spot a mistake

lightsprint09:
So I checked the tests again. They work fine.
I added 2 test cases to test JSON/multEmit without Buffer/Binary data. They both pass. 
Other tests with Binary still fail. 

I think this proves that the binary data causes the tests to fail. The tests should be valid

nuclearace:
Okay I see what's going on

nuclearace:
https://github.com/socketio/socket.io-client-swift/commit/150e26aefd8cc8b5b2bbfba8311bfd9862bb1927 should fix things. I noticed some tests still fail when be run all at once. But they pass when run individually.

lightsprint09:
Wow! Good job. All tests are passing

nuclearace:
No idea.

nuclearace:
Looks fine to merge to me. Is there anything else you wanted to do?

lightsprint09:
Not at the moment.

I will open a new PR if I find something to improve

PR #141, opened 2015-08-10
Api improvements

lightsprint09:
- Better API by making the socket status more readable and type-safe.
- Clean internal data representation by using ```public private(set) var x``` instead of prefixed variables and computed properties
- Convenience method for joining a namespace
- using more swift style ```guard```and ``ìf let x = x``` 

Sorry for the wired merge commits.


nuclearace:
My only concern is that the new enum isn't exposed to Objective-C. Other than that it looks good.

lightsprint09:
should be done with this https://github.com/lightsprint09/socket.io-client-swift/commit/25464ffb9de4a80c15b4c60eb0ff858c60b4bace

PR #156, opened 2015-09-01
Travis ci test

lightsprint09:
Tests are running on travis. Currently they all fail because it isn't possible to run a local node/socket.io server on travis while executing the tests.
My idea is to extract the server into another repo so it can be easily deployed to Heroku.

Do anyone have a better idea or an server to host the test server on?

Current Travis Build: https://travis-ci.org/lightsprint09/socket.io-client-swift


.travis.yml
```diff
@@ -0,0 +1,5 @@
+language: objective-c
```
nuclearace: Is there no swift option yet?

nuclearace:
I'll talk to @rauchg and see about what we can do for travis-ci testing. I know at one point it was turned on for this repo.

.travis.yml
lightsprint09: not not an explicit swift option.
But it works fine

lightsprint09:
Ok. I think turning travis back shouldn't be a problem.

I think getting a test server working is harder. Any idea which is better than heroku

lightsprint09:
This was not meant to be merged at this state. Tests are still failing on Travis and the test server problem is not solved yet. 

This breaks nothing. But if possible revert it. 

We need some kind of server to run the node/socket on to test the swift code against. As I mentioned before this could be a heroku server or something coustom. 

I want to hear your opinion about the server first before I will  implement it

nuclearace:
Would it be possible to add a before_script that launches the node server? I know that the travis environment has node installed

lightsprint09:
I don't think this is possible because Travis only provides a container runtime. This is of course iOS. I don't think it is possible to run code from a foreign container runtime

nuclearace:
Well if that's the case, would it make sense to just stand up a test server that the tests point to?

lightsprint09:
Yeah. Heroku is free but start up time is about 30 seconds. This slows tests down but should work. Another possibility is to host it somewhere els where the server stays always on

nuclearace:
Yeah, I have a VPS that I use. The only thing is that if the tests change I'd have to manually update the test server. Or find a way to auto-update

lightsprint09:
We could use github web hooks to automatically deploy new tests. Or something that works out of the box

nuclearace:
I think temporarily I'll just manually update the server

nuclearace:
So, it looks travis is running into some kind of problem

```
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'API violation - multiple calls made to -[XCTestExpectation fulfill] for connect.'
```

nuclearace:
Eck, it seems to be the same problem of sometimes the test pass and other times it fails

nuclearace:
https://travis-ci.org/socketio/socket.io-client-swift/builds/78425251

lightsprint09:
Do the tests pass locally ?

nuclearace:
Yeah when run locally all the tests pass

nuclearace:
The most common thing seems to be an error dealing with multiple calls to -[XCTestExpectation fulfill] like I posted above

lightsprint09:
Is it normal that the debugger stops here
screenshot

nuclearace:
If it's getting there that means either the client got released or something called stopPolling, which gets called in close.

lightsprint09:
Is this an intended behavior?
client is not nil and reason is "canceled"

nuclearace:
It looks like from your picture that something has called `close` on the engine. In which case it's normal to get to this check.

nuclearace:
Although I believe that this check is from an earlier version of the client where the engine was only added once. I think it could probably be taken out

lightsprint09:
I just found that travis does not support outside internet connection...
I´ll contact them for solution

nuclearace:
I guess the _right_ thing to do would be to mock a socket.io server. But I really don't feel writing a mock server.

lightsprint09:
year... but this would be a lot of work

lightsprint09:
I was wrong.

Travis support said it is indeed possible to reach external services. I take another look into in

PR #158, opened 2015-09-03
Fix travis ci tests

lightsprint09:
Sorry for all the trouble. This should fix travis ci.

All tests now use the same socketClient instance. I think opening 56x a connection was not the best way to to test.


nuclearace:
Nice, I thought it might have something to do with too many connections. This also speeds up the tests significantly.

PR #159, opened 2015-09-04
Fix secure option issue

brightsider:


nuclearace:
Is this an issue though? Specifying https should be enough

brightsider:
Then why options contains secure if this doesn't use?

nuclearace:
I don't see `secure` listed on the README for options

brightsider:
It was previously in the README, and there were no instructions for migration. May be worth add in the README that the need to specify the address with https?

nuclearace:
I'll add something to the README

PR #161, opened 2015-09-04
Logger +  Missing Test code

lightsprint09:
Simplify the Logger API by removing the requirment to implement LoggerClient to use the logger.

Added some code that satisfies the right namespace in wich the tests are executed

I going to test more of the code in the future

### DIFF
--- a/SocketIOClientSwift/SocketLogger.swift
+++ b/SocketIOClientSwift/SocketLogger.swift
@@ -26,43 +26,34 @@ import Foundation
 
 var Logger: SocketLogger = DefaultSocketLogger()
 
-public protocol SocketLogClient {
-    /// The type of object being logged
-    var logType: String {get}
-}
-
 public protocol SocketLogger {
     /// Whether to log or not
     var log: Bool {get set}
     
     /// Normal log messages
-    func log(message: String, client: SocketLogClient, altType: String?, args: AnyObject...)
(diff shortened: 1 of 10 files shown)


SocketIOClientSwift/SocketLogger.swift
```diff
@@ -26,43 +26,34 @@ import Foundation
...
+    }
+    
+    private func abstractLog(type:String, message: String, client: AnyObject, altType: String?, args:Array<AnyObject>) {
+        guard log else { return }
         
         let newArgs = args.map {arg -> CVarArgType in String(arg)}
         let replaced = String(format: message, arguments: newArgs)
-        
-        NSLog("ERROR %@: %@", altType ?? client.logType, replaced)
+        let sourceString =  NSStringFromClass(object_getClass(client))
```
nuclearace: I don't like using Objective-C API's like this in Swift. I would prefer to keep the requirement of conforming to a specific protocol to get the name of the type. Or ditch it entirely and just change altType to type

nuclearace:
I'll merge this, but I'm going to make a few changes.

SocketIOClientSwift/SocketLogger.swift
lightsprint09: Ok I understand that.

I would remove the alt type. It is a bit confusing

PR #164, opened 2015-09-07
More tests

lightsprint09:
This tests SocketParser and makes some code more readable.

It introduces a new class called GenericParser which handles reading strings.

### DIFF
--- a/GenericSocketParser.swift
+++ b/GenericSocketParser.swift
@@ -45,4 +45,8 @@ struct GenericParser {
         
         return subString.substringToIndex(foundRange.location)
     }
+    
+    mutating func readUntilEnd() ->String {
+        return read(messageCharacters.count - currentIndex)!
+    }
 }
--- a/GenericSocketParser.swift
+++ b/GenericSocketParser.swift
@@ -1,0 +1,48 @@
+//
+//  GenericSocketParser.swift
+//  Socket.IO-Client-Swift
+//
+//  Created by Lukas Schmidt on 07.09.15.
+//
+//
+
+import Foundation
+
+struct GenericParser {
+    let message: String
+    var currentIndex:Int
+    var messageCharacters: Array<Character> {
+        get {
+            return Array(message.characters)
+        }
+    }
+    var currentCharacter: String? {
+        get{
+            if currentIndex >= messageCharacters.count {
+                return nil
+            }
+            return String(messageCharacters[currentIndex])
+        }
+    }
+    
+    mutating func read(characterLength:Int) -> String? {
+        let startIndex = message.startIndex.advancedBy(currentIndex)
+        let range = Range<String.Index>(start: startIndex, end: startIndex.advancedBy(characterLength))
+        currentIndex = currentIndex + characterLength
+        
+        return message.substringWithRange(range)
+    }
+    
+    mutating func readUntilStringOccurence(string:String) -> String? {
+        let startIndex = message.startIndex.advancedBy(currentIndex)
+        let range = Range<String.Index>(start: startIndex, end: message.endIndex)
+        let subString = message.substringWithRange(range) as NSString
+        let foundRange = subString.rangeOfString(string)
+        if foundRange.location == Int.max {
+            return nil
+        }
+        currentIndex = currentIndex + foundRange.location
+        
+        return subString.substringToIndex(foundRange.location)
+    }
+}
--- a/GenericSocketParser.swift
+++ b/GenericSocketParser.swift
@@ -8,10 +8,10 @@
 
 import Foundation
 
-struct GenericParser {
+struct GenericSocketParser {
     let message: String
-    var currentIndex:Int
-    var messageCharacters: Array<Character> {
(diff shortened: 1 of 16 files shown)


nuclearace:
Couple of things:
I'm not really a fan of guard. I think just using ifs is fine.
I also do not like using Swift's error handling unless absolutely needed. Use optional if you can.

SocketIOClientSwift/SocketParser.swift
```diff
@@ -61,169 +60,119 @@ class SocketParser {
...
-        
-        socket.handleEvent(p.event, data: p.args,
-            isInternalMessage: false, wantsAck: p.id)
-    }
-    
     // Translation of socket.io-client#decodeString
-    static func parseString(str: String) -> SocketPacket? {
-        let arr = Array(str.characters)
-        let type = String(arr[0])
+    static func parseString(str: String) throws -> SocketPacket {
```
nuclearace: I would much prefer that this be kept an optional return type. Or if Swift ever introduces an Either type, that would be better

SocketIOClientSwift/SocketParser.swift
```diff
@@ -61,169 +60,119 @@ class SocketParser {
...
+        var parser = GenericParser(message: str, currentIndex: 0)
+        guard let typeString = parser.read(1), let type = SocketPacket.PacketType(str: typeString) else {
```
nuclearace: This is much too condensed. It serves it better if it's split into more lines.

nuclearace:
Also, I applaud you for trying to use my earlier style system, but I've decided to go with the flow and use a space after `:`

GenericSocketParser.swift
```diff
@@ -0,0 +1,52 @@
+//
+//  GenericSocketParser.swift
```
nuclearace: Should be named SocketGenericParser just to be consistent.

GenericSocketParser.swift
```diff
@@ -0,0 +1,52 @@
...
+//  Created by Lukas Schmidt on 07.09.15.
+//
+//
+
+import Foundation
+
+struct GenericParser {
+    let message: String
+    var currentIndex:Int
+    var messageCharacters: Array<Character> {
```
nuclearace: [Character] is the style I like to use

lightsprint09:
I think guard does really well when showing exits in a method -  it is more like a keyword

Why not using try/catch. It is a create mechanism for errors and using optional for errors is much less expressiv. 

I will go over the code and improve the style

GenericSocketParser.swift
lightsprint09: This is a good name. I was not sure how to name it

GenericSocketParser.swift
lightsprint09: ok

nuclearace:
I just don't see the need for guard, especially in these use cases. An if seems just as effective.

As for try/catch... It's more because I don't like the way Swift does it. While I understand the reason why you need to do `do {} catch` I just think it's ugly, whereas using optionals, while it means you don't get the specific reason for the failure, is better. At least in this case. Since we don't really care about a specific reason for failure.

And like I said, I really wish Swift would introduce Haskell's/Scala's `Either` type, or Rust's `Result` type. It would make exception handling prettier imo.

http://www.scala-lang.org/api/2.9.3/scala/Either.html

lightsprint09:
I don't get the point of theses `Either` types.
If you throw an Error you can with information attached it is nearly the same?

Would it be ok to stay with try/catch or should I remove it in favor of optionals

nuclearace:
The point of Either/Result is you don't need special error handling, you can just pattern match on Left and Right. Left being used for an error message string, and right for the successful value.

nuclearace:
I would prefer if we stayed with optionals.

nuclearace:
Also, here's an example of Either in Swift.

```swift
enum Either<E, T> {
    case Left(E)
    case Right(T)
}

func testEither(n: Int) -> Either<String, Int> {
    if n < 0 {
        return Either.Left("Less than 0")
    } else {
        return Either.Right(n + 4)
    }
}

let a = testEither(0)

switch a {
case Either.Left(let error):
    print("error: \(error)")
case Either.Right(let value):
    print("Success: \(value)")
}
```

lightsprint09:
Ok, fixed the classname

I will stay with optionals because is natural in swift. I won't implement the Either type because it feels not natural and people are not familiar with this. 

In my opinion native language features should be used in favor of other reimplementations because people are familiar with them

nuclearace:
Yeah, I wasn't suggesting implementing Either in Swift for the library. I suspect it'll be added to Swift later. Since it's a major feature of functional languages. And Swift draws a lot of influence from functional programming.

lightsprint09:
After reading your code example I got the idea of the `Either` type. It looks quite interesting, but I would not use it for error handling

nuclearace:
This looks good to me. Is it ready for merging? Or was there something else you wanted to do

lightsprint09:
ready :+1:

PR #96, opened 2015-06-09
Swift 2

nuclearace:
This will be merged immediately when Swift 2 is officially released.

PR #170, opened 2015-09-11
Code improvements to Swift 2 branch

NachoSoto:
I made a bunch of comments in #96, and here's my proposed solution to them, plus a few other improvements I've found along the way. Let me know what you think.
All tests are still passing.

### Changes:
- `SSLCert` is now `private`.
- `WebSockets`: removed unnecessary `weak`s and simplified code.
- `SwiftRegex` does not need to be exposed.
- `SwiftRegex`: failing with `fatalError`.
- `SwiftRegex`: changed static method to `static` to avoid unnecessary dynamic dispatching.
- `SocketPacket`: using `Logger` instead of `print`.
- `SocketIOClient`: simplified conditional.
- `SocketIOClient`: remove unnecessary `optional`.
- `SocketIOClient`: removed unnecessary `weak self`.
- `SocketIOClient`: added assertion to check `timeoutAfter` is not negative.
- `SocketIOClient`: making a few fields `private`.
- Simplified `SocketEngine.swift`.

### DIFF
--- a/SocketIOClientSwift/SocketEngine.swift
+++ b/SocketIOClientSwift/SocketEngine.swift
@@ -350,9 +350,11 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
     // We had packets waiting for send when we upgraded
     // Send them raw
     private func flushWaitingForPostToWebSocket() {
-        for msg in postWait {
-            ws?.writeString(msg)
-        }
+		if let ws = ws {
+			for msg in postWait {
+				ws.writeString(msg)
+			}
+		}
 
         postWait.removeAll(keepCapacity: true)
     }
@@ -495,7 +497,7 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
         var n = 0
         var msg = ""
 
-        func testLength(length:String, inout n:Int) -> Bool {
+        func testLength(length: String, inout n: Int) -> Bool {
             if let num = Int(length) {
                 n = num
                 return false
@@ -525,8 +527,8 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
 
                 if msg.characters.count != 0 {
                     // Be sure to capture the value of the msg
-                    dispatch_async(handleQueue) {[weak self, msg] in
-                        self?.parseEngineMessage(msg, fromPolling: true)
+					dispatch_async(handleQueue) {
+                        self.parseEngineMessage(msg, fromPolling: true)
                     }
                 }
 
@@ -549,8 +551,8 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
 
         let type = PacketType(str: (message["^(\\d)"].groups()?[1]) ?? "") ?? {
             self.checkIfMessageIsBase64Binary(message)
-            return PacketType.Noop
-            }()
+            return .Noop
+		}()
 
         switch type {
         case PacketType.Message:
@@ -608,13 +610,11 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
 
             postWait.append(strMsg)
 
-            if let datas = datas {
-                for data in datas {
-                    let (_, b64Data) = createBinaryDataForSend(data)
+			for data in datas ?? [] {
+				let (_, b64Data) = createBinaryDataForSend(data)
 
-                    postWait.append(b64Data!)
-                }
-            }
+				postWait.append(b64Data!)
+			}
 
             if !waitingForPost {
                 flushWaitingForPost()
@@ -629,30 +629,26 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
 
             ws?.writeString("\(type.rawValue)\(str)")
 
-            if let datas = datas {
-                for data in datas {
-                    let (data, _) = createBinaryDataForSend(data)
-                    if data != nil {
-                        ws?.writeData(data!)
-                    }
-                }
-            }
+			for data in datas ?? [] {
+				let (data, _) = createBinaryDataForSend(data)
+				if data != nil {
+					ws?.writeData(data!)
+				}
+			}
     }
 
     // Starts the ping timer
     private func startPingTimer() {
-        guard pingInterval != nil else {
-            return
-        }
+		if let pingInterval = pingInterval {
+			pingTimer?.invalidate()
+			pingTimer = nil
 
-        pingTimer?.invalidate()
-        dispatch_async(dispatch_get_main_queue()) {[weak self] in
-            if let this = self {
-                this.pingTimer = NSTimer.scheduledTimerWithTimeInterval(this.pingInterval!, target: this,
-                    selector: Selector("sendPing"), userInfo: nil, repeats: true)
-            }
-        }
-    }
+			dispatch_async(dispatch_get_main_queue()) {
+				self.pingTimer = NSTimer.scheduledTimerWithTimeInterval(pingInterval, target: self,
+					selector: Selector("sendPing"), userInfo: nil, repeats: true)
+			}
+		}
+	}
 
     func stopPolling() {
         session.invalidateAndCancel()
@@ -672,16 +668,16 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
     Write a message, independent of transport.
     */
     public func write(msg: String, withType type: PacketType, withData data: [NSData]?) {
-        dispatch_async(emitQueue) {[weak self] in
-            if let this = self where this.connected {
-                if this.websocket {
-                    Logger.log("Writing ws: %@ has data: %@", type: this.logType, args: msg,
+		dispatch_async(emitQueue) {
+            if self.connected {
+                if self.websocket {
+                    Logger.log("Writing ws: %@ has data: %@", type: self.logType, args: msg,
                         data == nil ? false : true)
-                    this.sendWebSocketMessage(msg, withType: type, datas: data)
+                    self.sendWebSocketMessage(msg, withType: type, datas: data)
                 } else {
-                    Logger.log("Writing poll: %@ has data: %@", type: this.logType, args: msg,
+                    Logger.log("Writing poll: %@ has data: %@", type: self.logType, args: msg,
                         data == nil ? false : true)
-                    this.sendPollMessage(msg, withType: type, datas: data)
+                    self.sendPollMessage(msg, withType: type, datas: data)
                 }
             }
         }
--- a/SocketIOClientSwift/SocketEngine.swift
+++ b/SocketIOClientSwift/SocketEngine.swift
@@ -24,7 +24,7 @@
 
 import Foundation
 
-public final class SocketEngine: NSObject, WebSocketDelegate {    
+public final class SocketEngine: NSObject, WebSocketDelegate {
     private typealias Probe = (msg: String, type: PacketType, data: [NSData]?)
     private typealias ProbeWaitQueue = [Probe]
 
@@ -350,11 +350,11 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
     // We had packets waiting for send when we upgraded
     // Send them raw
     private func flushWaitingForPostToWebSocket() {
-		if let ws = ws {
-			for msg in postWait {
-				ws.writeString(msg)
-			}
-		}
+        if let ws = ws {
+            for msg in postWait {
+                ws.writeString(msg)
+            }
+        }
 
         postWait.removeAll(keepCapacity: true)
     }
@@ -527,7 +527,7 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
 
                 if msg.characters.count != 0 {
                     // Be sure to capture the value of the msg
-					dispatch_async(handleQueue) {
+                    dispatch_async(handleQueue) {
                         self.parseEngineMessage(msg, fromPolling: true)
                     }
                 }
@@ -552,7 +552,7 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
         let type = PacketType(str: (message["^(\\d)"].groups()?[1]) ?? "") ?? {
             self.checkIfMessageIsBase64Binary(message)
             return .Noop
-		}()
+            }()
 
         switch type {
         case PacketType.Message:
@@ -610,11 +610,11 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
 
             postWait.append(strMsg)
 
-			for data in datas ?? [] {
-				let (_, b64Data) = createBinaryDataForSend(data)
+            for data in datas ?? [] {
+                let (_, b64Data) = createBinaryDataForSend(data)
 
-				postWait.append(b64Data!)
-			}
+                postWait.append(b64Data!)
+            }
 
             if !waitingForPost {
                 flushWaitingForPost()
@@ -629,26 +629,26 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
 
             ws?.writeString("\(type.rawValue)\(str)")
 
-			for data in datas ?? [] {
-				let (data, _) = createBinaryDataForSend(data)
-				if data != nil {
-					ws?.writeData(data!)
-				}
-			}
+            for data in datas ?? [] {
+                let (data, _) = createBinaryDataForSend(data)
+                if data != nil {
+                    ws?.writeData(data!)
+                }
+            }
     }
 
     // Starts the ping timer
     private func startPingTimer() {
-		if let pingInterval = pingInterval {
-			pingTimer?.invalidate()
-			pingTimer = nil
+        if let pingInterval = pingInterval {
+            pingTimer?.invalidate()
+            pingTimer = nil
 
-			dispatch_async(dispatch_get_main_queue()) {
-				self.pingTimer = NSTimer.scheduledTimerWithTimeInterval(pingInterval, target: self,
-					selector: Selector("sendPing"), userInfo: nil, repeats: true)
-			}
-		}
-	}
+            dispatch_async(dispatch_get_main_queue()) {
+                self.pingTimer = NSTimer.scheduledTimerWithTimeInterval(pingInterval, target: self,
+                    selector: Selector("sendPing"), userInfo: nil, repeats: true)
+            }
+        }
+    }
 
     func stopPolling() {
         session.invalidateAndCancel()
@@ -668,7 +668,7 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
     Write a message, independent of transport.
     */
     public func write(msg: String, withType type: PacketType, withData data: [NSData]?) {
-		dispatch_async(emitQueue) {
+        dispatch_async(emitQueue) {
             if self.connected {
                 if self.websocket {
                     Logger.log("Writing ws: %@ has data: %@", type: self.logType, args: msg,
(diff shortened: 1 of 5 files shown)


SocketIOClientSwift/SocketEngine.swift
```diff
@@ -350,9 +350,11 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
     // We had packets waiting for send when we upgraded
     // Send them raw
     private func flushWaitingForPostToWebSocket() {
-        for msg in postWait {
-            ws?.writeString(msg)
-        }
+		if let ws = ws {
```
NachoSoto: This avoids looping through `postWait` if `ws` is `nil`.

SocketIOClientSwift/SocketEngine.swift
```diff
@@ -629,26 +629,22 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
...
                 }
             }
     }
 
     // Starts the ping timer
     private func startPingTimer() {
-        guard pingInterval != nil else {
-            return
-        }
+        if let pingInterval = pingInterval {
```
lightsprint09: I would use` guard let  pingInterval = pingInterval else { return }`

hankbao:
The `weak` remained in `WebSocket.connect()` method actually causes a nasty race condition problem. 

The WebSocket instance within SocketEngine is released in `createWebsocket` but may still be accessed in a global concurrent queue in `WebSocket.connect()`.

`let weakSelf = self` will try to unwrap and retain a deallocated object and crash the app.

SocketIOClientSwift/SocketEngine.swift
```diff
@@ -350,8 +350,10 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
     // We had packets waiting for send when we upgraded
     // Send them raw
     private func flushWaitingForPostToWebSocket() {
-        for msg in postWait {
-            ws?.writeString(msg)
+        if let ws = ws {
```
nuclearace: There's no need to break this out. `ws?.writeString(msg)` won't crash if ws is nil

SocketIOClientSwift/SocketEngine.swift
nuclearace: I actually prefer using if let

SocketIOClientSwift/SocketEngine.swift
```diff
@@ -525,8 +527,8 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
 
                 if msg.characters.count != 0 {
                     // Be sure to capture the value of the msg
-                    dispatch_async(handleQueue) {[weak self, msg] in
```
nuclearace: Is there a reason you're removing the weak self? Generally when you dispatch async you don't want to create a strong reference

SocketIOClientSwift/SocketEngine.swift
```diff
@@ -672,16 +668,16 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
     Write a message, independent of transport.
     */
     public func write(msg: String, withType type: PacketType, withData data: [NSData]?) {
-        dispatch_async(emitQueue) {[weak self] in
```
nuclearace: I would keep this weak, there's no need to keep a strong reference.

SocketIOClientSwift/SocketIOClient.swift
```diff
@@ -148,28 +153,30 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
     */
     public func connect(timeoutAfter timeoutAfter:Int,
         withTimeoutHandler handler:(() -> Void)?) {
-            guard status != SocketIOClientStatus.Connected else {
+            assert(timeoutAfter >= 0, "Invalid timeout: \(timeoutAfter)")
```
nuclearace: Don't use assert

SocketIOClientSwift/SocketIOClient.swift
```diff
@@ -148,28 +153,30 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
...
-            engine?.open(connectParams)
+            addEngine().open(connectParams)
             
             guard timeoutAfter != 0 else {
                 return
             }
             
             let time = dispatch_time(DISPATCH_TIME_NOW, Int64(timeoutAfter) * Int64(NSEC_PER_SEC))
-            
-            dispatch_after(time, dispatch_get_main_queue()) {[weak self] in
```
nuclearace: Again, I don't think it's good to keep strong references like this

SocketIOClientSwift/SocketIOClient.swift
```diff
@@ -181,15 +188,15 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
             if let this = self {
                 this.ackHandlers.addAck(ack, callback: callback)
                 
-                dispatch_async(this.emitQueue) {[weak this] in
```
nuclearace: There's no reason this should change to a strong reference

SocketIOClientSwift/SocketIOClient.swift
```diff
@@ -181,15 +188,15 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
...
+                dispatch_async(this.emitQueue) {
+                    this._emit(items, ack: ack)
                 }
                 
                 if timeout != 0 {
                     let time = dispatch_time(DISPATCH_TIME_NOW, Int64(timeout * NSEC_PER_SEC))
                     
-                    dispatch_after(time, dispatch_get_main_queue()) {[weak this] in
-                        this?.ackHandlers.timeoutAck(ack)
+                    dispatch_after(time, dispatch_get_main_queue()) {
```
nuclearace: Should be a weak `this`

SocketIOClientSwift/SocketIOClient.swift
```diff
@@ -249,12 +256,12 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
...
     public func emit(event: String, withItems items: [AnyObject]) {
-        guard status == SocketIOClientStatus.Connected else {
+        guard status == .Connected else {
             return
         }
-        
-        dispatch_async(emitQueue) {[weak self] in
-            self?._emit([event] + items)
+
+        dispatch_async(emitQueue) {
```
nuclearace: `weak self` is fine here

SocketIOClientSwift/SocketIOClient.swift
```diff
@@ -292,17 +299,17 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
     
     // If the server wants to know that the client received data
     func emitAck(ack: Int, withItems items: [AnyObject]) {
-        dispatch_async(emitQueue) {[weak self] in
```
nuclearace: There's not really a reason for this to be strong.

SocketIOClientSwift/SocketIOClient.swift
```diff
@@ -334,22 +341,21 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
...
                 return
             }
-            // println("Should do event: \(event) with data: \(data)")
             
             Logger.log("Handling event: %@ with data: %@", type: logType, args: event, data ?? "")
             
             if anyHandler != nil {
-                dispatch_async(handleQueue) {[weak self] in
-                    self?.anyHandler?(SocketAnyEvent(event: event, items: data))
+                dispatch_async(handleQueue) {
```
nuclearace: Should be `weak self`

SocketIOClientSwift/SocketIOClient.swift
```diff
@@ -334,22 +341,21 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
...
+                    self.anyHandler?(SocketAnyEvent(event: event, items: data))
                 }
             }
             
             for handler in handlers where handler.event == event {
                 if let ack = ack {
-                    dispatch_async(handleQueue) {[weak self] in
+                    dispatch_async(handleQueue) {
```
nuclearace: Should be `weak self`

SocketIOClientSwift/SocketIOClient.swift
```diff
@@ -441,18 +447,14 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
     }
     
     public func parseSocketMessage(msg: String) {
-        dispatch_async(handleQueue) {[weak self] in
-            if let this = self {
-                SocketParser.parseSocketMessage(msg, socket: this)
-            }
+        dispatch_async(handleQueue) {
```
nuclearace: Should be `weak self`

SocketIOClientSwift/SocketIOClient.swift
```diff
@@ -441,18 +447,14 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
...
+            SocketParser.parseSocketMessage(msg, socket: self)
         }
     }
     
     public func parseBinaryData(data: NSData) {
-        dispatch_async(handleQueue) {[weak self] in
-            if let this = self {
-                SocketParser.parseBinaryData(data, socket: this)
-            }
+        dispatch_async(handleQueue) {
```
nuclearace: Should be `weak self`

SocketIOClientSwift/SocketIOClient.swift
```diff
@@ -468,19 +470,17 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
...
             
-            status = SocketIOClientStatus.Reconnecting
+            status = .Reconnecting
             
-            dispatch_async(dispatch_get_main_queue()) {[weak self] in
-                if let this = self {
-                    this.reconnectTimer = NSTimer.scheduledTimerWithTimeInterval(Double(this.reconnectWait),
-                        target: this, selector: "_tryReconnect", userInfo: nil, repeats: true)
-                }
+            dispatch_async(dispatch_get_main_queue()) {
```
nuclearace: Should be `weak self`

SocketIOClientSwift/SocketPacket.swift
```diff
@@ -28,6 +28,8 @@ struct SocketPacket {
     private let placeholders: Int
     private var currentPlace = 0
 
+	private static let logType = "SocketPacket"
```
nuclearace: Indentation is a little wonky in this file

SocketIOClientSwift/SocketEngine.swift
lightsprint09: but you win one indentation. It makes sense in methods with long single lines of code

SocketIOClientSwift/WebSocket.swift
```diff
@@ -378,21 +378,17 @@ public class WebSocket : NSObject, NSStreamDelegate {
         }
         if totalSize > 0 {
             if validateResponse(buffer, bufferLen: totalSize) {
-                dispatch_async(queue, { [weak self] in
```
nuclearace: I believe these weak references are on purpose, otherwise there's a case where the WebSocket won't be released.

nuclearace:
Two things. Generally I take the approach that unless there's an explicit reason to perform a strong capture, always perform a weak capture. This way you avoid crazy cases where you've got a strong reference that's keeping your object alive.

Also, the weak references in the WebSocket are intentional. https://github.com/daltoniam/Starscream/commit/889d2880eb4585373b1bc3907574399347d440ff

hankbao:
@nuclearace There is a race condition in `WebSocket.connect()`. Please see my [comment](https://github.com/socketio/socket.io-client-swift/pull/170#issuecomment-139485493) above.

I would like to open an issue about that if you think that is appropriate.

nuclearace:
@hankbao I thought you fixed that in the thing I merged yesterday? Is it still present in the default [WebSocket](https://github.com/daltoniam/Starscream)? I generally try to back-patch those sort of things. If it's still an issue there I'd make an issue there.

hankbao:
@nuclearace IMO, the multithreads race condition in `connect` method is different from the use-after-release problem caused by a unowned object and that is not fixed by weak-referencing self. After turned on forceWebsocket, it's now easy to reproduce the problem on iOS 9 GM if you connect-close-connect quickly.

nuclearace:
You're welcome to provide a fix, I'm having trouble reproducing it on the swift-2 branch. But I do most of my testing on OS X.

lightsprint09:
@hankbao it would be great to create a test case to test against automatically

hankbao:
@nuclearace @lightsprint09 OK. But I'm quite busy currently and it's almost at midnight in China. I'd rather work on this issue tomorrow.

lightsprint09:
I think we are all in different time zones :+1:

nuclearace:
@hankbao That's fine. I don't think this is really a major issue

SocketIOClientSwift/SocketEngine.swift
NachoSoto: I prefer to avoid early returns that can be problematic if you don't notice them in the future. But I understand that's an stylistic choice (unlike some other things in this PR), so I'll chose whatever you guys prefer.

SocketIOClientSwift/SocketIOClient.swift
NachoSoto: Why not? It doesn't make any sense to pass a negative `timeout`. What should be the expected behavior? The `dispatch_after` would be invoked right away so it would timeout right away.

SocketIOClientSwift/SocketPacket.swift
NachoSoto: Sorry! I'll fix

SocketIOClientSwift/WebSocket.swift
NachoSoto: This strong reference is only kept for a short period of time (while this block is invoked). This **won't** cause a retain cycle. I'll explain more separately though.

SocketIOClientSwift/SocketEngine.swift
NachoSoto: This is a small optimization: there's no need to iterate through `postWait` if `ws` is `nil` anyway.

SocketIOClientSwift/SocketIOClient.swift
lightsprint09: would it make sense to change the parameter type from `Int` to `UInt`

NachoSoto:
> @nuclearace IMO, the multithreads race condition in connect method is different from the use-after-release problem caused by a unowned object and that is not fixed by weak-referencing self. After turned on forceWebsocket, it's now easy to reproduce the problem on iOS 9 GM if you connect-close-connect quick

That's right, there's a very big difference between unowned and weak. In that sense, something like [this commit](https://github.com/daltoniam/Starscream/commit/889d2880eb4585373b1bc3907574399347d440ff) is right because `unowned` can crash the app.

> let weakSelf = self will try to unwrap and retain a deallocated object and eventually crashes the app.

I'm sorry, but that's completely wrong. if `self` is a weak reference, doing `if let weakSelf = self` won't crash the app. `self` will have been nil-ed out (because it's `weak`, not `unowned`), which means that doing `if-let` simply defines `weakSelf` **only if** it's not `nil`. That's completely say.

>Two things. Generally I take the approach that unless there's an explicit reason to perform a strong capture, always perform a weak capture. This way you avoid crazy cases where you've got a strong reference that's keeping your object alive.

Using `weak` is useful for places where you want to avoid a retain cycle, like [here](https://github.com/socketio/socket.io-client-swift/blob/swift-2/SocketIOClientSwift/SocketIOClient.swift#L180).

But all the ones I've removed in this PR are unnecessary:
- `dispatch_*` is guaranteed to be invoked some time in the future, which means that the cycle will always be broken.
- `weak` introduces an unnecessary performance hit: there needs to be extra work to monitor references to them `nil-out`.
- `weak` introduces unnecessary complexity: you need to consider in each code path what the right behavior should be if `self` is dealloced. We can **simplify** the code by avoiding this.
- `weak` introduces more code than needed: look at how much code I removed: `weak self`, `guard`, etc. Code becomes a lot simpler, while keeping the same behavior.

NachoSoto:
Thanks for the comments guys!

SocketIOClientSwift/SocketIOClient.swift
NachoSoto: Apple is against unsigned types (they've almost removed them from all their APIs), but I do agree with that solution, specially since Swift guards against overflows.

nuclearace:
>dispatch_* is guaranteed to be invoked some time in the future, which means that the cycle will always be broken.

I wasn't aware of that

SocketIOClientSwift/SocketIOClient.swift
nuclearace: I'm also against unsigned types. Just use checks or treat negatives as 0

nuclearace:
I would prefer to keep the WebSockets library as is. If you think the `weak`s are uneeded, make an issue on that [repo](https://github.com/daltoniam/Starscream) I generally try to keep up that

NachoSoto:
Why would an issue be more helpful? I've already done the work for you.

Is there a particular reason you disagree with this other than wanting to keep `weak` because you don't understand it?

What about the rest of changes?

nuclearace:
Taking the `weak`s out is fine if what you say about dispatch_* always executing the closure is true. I just wasn't sure about the WebSocket one, since I didn't write it.

NachoSoto:
Cool. Which one aren't you sure about? Maybe one of them is actually needed.

SocketIOClientSwift/SocketEngine.swift
nuclearace: Is the optimization really worth it? I'd prefer to keep as is, unless you've determined this is a bottleneck

NachoSoto:
It's not *just* for the optimization, it's mostly simplifying the code to make it easier to maintain :)

NachoSoto:
And because it means we no longer have to think about what should happen if self is nil everywhere.

nuclearace:
The ones in https://github.com/daltoniam/Starscream/commit/889d2880eb4585373b1bc3907574399347d440ff Not sure if you removed some of these.

nuclearace:
And fyi, I'm all for taking the `weak`s out, I just wasn't sure if there's going to be some edge case where there's a strong retain that's keeping the client from being released.

SocketIOClientSwift/SocketEngine.swift
```diff
@@ -495,7 +497,7 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
         var n = 0
         var msg = ""
 
-        func testLength(length:String, inout n:Int) -> Bool {
+        func testLength(length: String, inout n: Int) -> Bool {
```
nuclearace: Also thanks, I keep finding places where I've forgotten to remove the old style.

NachoSoto:
> The ones in daltoniam/Starscream@889d288 Not sure if you removed some of these.

I haven't :)

> And fyi, I'm all for taking the weaks out, I just wasn't sure if there's going to be some edge case where there's a strong retain that's keeping the client from being released.

I studied each of them very rigorously (I didn't just got rid of all `weak`s). I only removed the ones where it's guaranteed that the cycle will be broken. In the case of `dispatch`, the block disappears (and with it, its `strong` `self` reference) as soon as it's invoked. It's impossible that that will create a cycle. There's actually never a cycle because `self` doesn't own the block.

NachoSoto:
Thanks :star:

PR #172, opened 2015-09-12
Added support for tvOS

NachoSoto:
## Changes:

- Duplicated target for both framework and tests.
- Set the right platform on new targets.
- Changed `tvOS` framework target product name to `SocketIO`.
- Changed targets `plist`s to use the existing ones, and removed the copies.
- Configured scheme to use the right test target.
- Shared scheme.
- Verified `tvOS` tests pass.

screen shot 2015-09-11 at 17 15 58

### DIFF
--- a/Socket.IO-Client-Swift.xcodeproj/project.pbxproj
+++ b/Socket.IO-Client-Swift.xcodeproj/project.pbxproj
@@ -12,6 +12,47 @@
 		572EF23D1B51F18A00EEBB58 /* SocketIO-Mac.h in Headers */ = {isa = PBXBuildFile; fileRef = 572EF23C1B51F18A00EEBB58 /* SocketIO-Mac.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		572EF2431B51F18A00EEBB58 /* SocketIO.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 572EF2381B51F18A00EEBB58 /* SocketIO.framework */; };
 		572EF24A1B51F18A00EEBB58 /* SocketIO_MacTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 572EF2491B51F18A00EEBB58 /* SocketIO_MacTests.swift */; };
+		57425F9C1BA3A46000BDAAC1 /* SocketGenericParser.swift in Sources */ = {isa = PBXBuildFile; fileRef = 74D765611B9F0D870028551C /* SocketGenericParser.swift */; };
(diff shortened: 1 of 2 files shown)


nuclearace:
Is this going to cause problems for Xcode 7 users? If so, this should probably be merged into an Xcode 7.1 branch

NachoSoto:
It won't. If you try to build the new framework you'll get this error:
screen shot 2015-09-11 at 18 25 34

But other than that, nothing is affected, so I think it's safe and simpler to integrate this already for anyone who wants to build this for the Apple TV.

If you prefer, though, we can leave it in a separate branch.

nuclearace:
As long as it doesn't affect the building of the other targets, I'm fine with putting it here.

NachoSoto:
:+1:

victorpimentel:
I'm sorry, but is this causing trouble to anyone using Carthage? If you create a sample project and add this to Cartfile:

```
github "socketio/socket.io-client-swift" "swift-2"
```

It fails when you try to install this framework using Carthage and Xcode 7.1:

```
→ carthage update
*** Fetching socket.io-client-swift
*** Checking out socket.io-client-swift at "60e014295517d9a5770b5743f81d1d59f6b89454"
*** xcodebuild output can be found in /var/folders/q_/75868zgn3d548f5t_n8r69_h0000gn/T/carthage-xcodebuild.OrIjZj.log
*** Building scheme "SocketIO-Mac" in Socket.IO-Client-Swift.xcodeproj
*** Building scheme "SocketIO-iOS" in Socket.IO-Client-Swift.xcodeproj
*** Building scheme "SocketIO-tvOS" in Socket.IO-Client-Swift.xcodeproj
Parse error: unexpected SDK key "appletvos"
```

Pointing the Cartfile to the changeset before this pull request was merged (17444dd), correctly generates the frameworks with no errors at all.

I don't know exactly how to solve this or if this is just a Carthage limitation and this does not happen with Cocoapods or any other dependency manager, I didn't investigate this much further. In any case, IMHO, now that we have a Xcode 7 GM we should have two branches, one for Xcode 7 and another one for Xcode 7.1, at least until 7.1 reaches the GM point.

NachoSoto:
@victorpimentel what you want to do is limit the platforms you want to build for. You should always do that, as I doubt you need to build for all 3 architectures, which is very time consuming :)

```
carthage update --platform ios
carthage update --platform macosx
```

Or with Carthage 0.9
```
carthage update --platform ios,macosx
```

PR #183, opened 2015-09-17
Adds SocketIOClient#once(event, callback)

KevinMartin:
When the specified event is emitted, the handler deregisters itself.

### DIFF
--- a/SocketIOClientSwift/SocketEventHandler.swift
+++ b/SocketIOClientSwift/SocketEventHandler.swift
@@ -35,6 +35,7 @@ private func emitAckCallbackObjectiveC(socket: SocketIOClient?, num: Int?)
 }
 
 struct SocketEventHandler {
+    let id = NSUUID()
     let event: String
     let callback: NormalCallback?
     let callBackObjectiveC: NormalCallbackObjectiveC?
--- a/SocketIOClientSwift/SocketIOClient.swift
+++ b/SocketIOClientSwift/SocketIOClient.swift
@@ -432,7 +432,7 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
         
         var handler: SocketEventHandler;
 
-        handler = SocketEventHandler(event: event) { data, ack in
+        handler = SocketEventHandler(event: event) { (data, ack: AckEmitter?) in
             self.handlers = ContiguousArray(self.handlers.filter { $0 != handler })
             callback?(data, ack)
         }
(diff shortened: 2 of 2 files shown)


SocketIOClientSwift/SocketIOClient.swift
```diff
@@ -425,6 +425,34 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
     }
     
     /**
+    Adds a single-use handler for an event.
+    */
+    public func once(event: String, callback: NormalCallback) {
+        Logger.log("Adding once handler for event: %@", type: logType, args: event)
+        
+        let handler = SocketEventHandler(event: event) { data: NSArray?, ack: AckEmitter? in
```
nuclearace: You shouldn't need type declarations here. 

```swift
let handler = SocketEventHandler(event: event) {data, ack in
```

Should be fine

SocketIOClientSwift/SocketIOClient.swift
```diff
@@ -425,6 +425,34 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
...
+            handlers = ContiguousArray(handlers.filter { $0 != handler })
```
nuclearace: Shouldn't this be a weak reference?

SocketIOClientSwift/SocketIOClient.swift
```diff
@@ -425,6 +425,34 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
...
+        handlers.append(handler)
+    }
+    
+    /**
+    Adds a single-use handler for an event.
+    */
+    public func once(event event: String, callback: NormalCallbackObjectiveC) {
+        Logger.log("Adding once handler for event: %@", type: logType, args: event)
+        
+        let handler = SocketEventHandler(event: event) { data: NSArray?, ack: AckEmitterObjectiveC? in
```
nuclearace: Seems to be a build error here

SocketIOClientSwift/SocketIOClient.swift
NachoSoto: No, `$0` won't outlive this declaration so there's no need to not make it `strong`.

nuclearace: I'm talking about referencing self.handlers inside of this entire block

KevinMartin:
All fixed! :+1:

SocketIOClientSwift/SocketEventHandler.swift
```diff
@@ -35,6 +35,7 @@ private func emitAckCallbackObjectiveC(socket: SocketIOClient?, num: Int?)
 }
 
 struct SocketEventHandler {
+    let id = NSUUID()
```
nuclearace: Why is this needed?

KevinMartin: You can't compare structs. I looked at other Swift-based event dispatchers that support removing specific handlers and they use id's like this to compare them.

lightsprint09:
It would be great to cover this functions with tests... In my opinion every new functions should be tested, before the PR is merged

PR #192, opened 2015-09-24
Either

nuclearace:
This changes the internal API to use an [Either type](https://en.wikipedia.org/wiki/Tagged_union) rather than optionals or tuples of optionals.

This allows for better error reporting and more concise code

### DIFF
--- a/SocketIOClientSwift/SocketEngine.swift
+++ b/SocketIOClientSwift/SocketEngine.swift
@@ -137,13 +137,13 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
 
             mutData.appendData(data)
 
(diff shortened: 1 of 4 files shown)


SocketIOClientSwift/SocketEngine.swift
```diff
@@ -129,7 +130,7 @@ public final class SocketEngine: NSObject, WebSocketDelegate {
         stopPolling()
     }
 
-    private func createBinaryDataForSend(data: NSData) -> (NSData?, String?) {
+    private func createBinaryDataForSend(data: NSData) -> Either<NSData, String> {
```
NachoSoto: Nice :)

PR #194, opened 2015-09-26
Test refactors

nuclearace:
- Moves back to xctool
- Separates the namespace socket from the regular socket. This will better expose issues in namespaces


nuclearace:
Ugh, I think I introduced that old test bug where sometimes the tests randomly fail.

nuclearace:
Probably the same problem we had before. But I think we need multiple sockets to test namespaces and polling stuff.

nuclearace:
@lightsprint09 do you have any suggestions with this? I'd like to try and separate the sockets based on what they're testing. But having 4 different sockets seems to be causing travis problems.

lightsprint09:
I didn't have time to take a look at it. Hopefully I find some time tomorrow

lightsprint09:
I would not kill the old tests
They are as important as the new ones

nuclearace:
Going to go ahead and merge this and continuously work on adding more tests

lightsprint09:
did you remove all the old tests

nuclearace:
Yeah, moving forward it'll be done more with mocking. Eventually I'm going to add some tests for the transport upgrading and connecting to namespaces. But the parsing stuff will mock tested. The client tests are pretty easy to mock out. But I think I'm going to need to refactor the engine to make easier to do testing on.

PR #199, opened 2015-10-06
Travis fix methods

lightsprint09:
Tests run fine, bu they fail sometimes

Don't merge this until I am sure that every tests passes everytime.

Few things I did:

- Move every test to the Mac-Target
- Polling-Tests inherit from AbstractTest and implement their own tests
-Eyery test owns a instance of a socketClient and don't uses the single shared static socket. 
-remove parallelization of test. Tests pass in about 2 seconds. There is no need to parallelize test, because parallelization is hard and causes hard to catch problems

My when the build passes I do a rebuild and then it fails...


nuclearace:
Another solution that would require a bit more work would be to cut out the server and instead test the output from the socket client to the engine and from the engine to the server, and vise versa. This would require some rewriting of some methods. Not sure if I want to go down that road just yet

lightsprint09:
This would more follow the unit test principles. I think it is an good idea but it come with a lot of work.

I would see our current approach more like an integration test. 
I tried the whole day to fix the test on travis. They run perfectly on the Mac even executed by xctool. Maybe there is an issue with the polling, because polling does fail often and I don't know why.
May you could take a look at the polling code

PR #206, opened 2015-10-14
Fixed incorrect code signing and bit code settings.

drekka:
When used in carthage I had failed builds because the iOS code signing was referring to a non-standard id. I then hit an issue where the OSX build was picking up a bit code setting from the project settings.

Have fixed both issues in this pull request.


Socket.IO-Client-Swift.xcodeproj/project.pbxproj
```diff
@@ -573,9 +570,6 @@
 		572EF2131B51F12F00EEBB58 /* Release */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				BITCODE_GENERATION_MODE = bitcode;
-				CODE_SIGN_IDENTITY = "Developer ID Application";
-				ENABLE_BITCODE = YES;
```
NachoSoto: This isn't defined in the targets, so you'll need to set it to `YES` explicitly for `iOS` and `tvOS`.

nuclearace:
Tbh my carthage is borked since I don't think they've updated it to work with OS X 10.11 and Swift 2.

nuclearace:
Is this still a problem in the latest version?

PR #211, opened 2015-10-17
Improve initial options

lightsprint09:
Use predefined keys in the dictionary to prevent typos

My initial intention was to remove key-value at all.
The initial options should be type save as well, like

ForceWebsocket(<expect Bool>)
CustomLogger(<expect Logger Implementation>)

But after all the solution should be easy to transfer to an [String: AnyObject]. Any ideas?


nuclearace:
I don't know if this a good approach. I don't think that typos in options is really that big of a problem.

lightsprint09:
I think it helps. And does not break anything

But a real type save solution would be even better

lightsprint09:
this commit partly proves it https://github.com/socketio/socket.io-client-swift/commit/5d867f4f6524f302d52fdc60c6907b6687dc8733 ;)

look at AbstractTest

nuclearace:
Heh, yeah. I've got another solution in mind though

lightsprint09:
Ok? What other solution?

nuclearace:
Similar to what you had, but using an enum.
```swift
public typealias SocketOptionsDictionary = [SocketIOClientOptions: AnyObject]

public enum SocketIOClientOptions: String {
    case ConnectParams = "connectParams"
    case Reconnects = "reconnect"
    case ReconnectAttempts = "reconnectAttempts"
    case ReconnectWait = "reconnectWait"
    case ForcePolling = "forcePolling"
    case ForceWebsockets = "forceWebsockets"
    case Nsp = "nsp"
    case Cookies = "cookies"
    case Log = "log"
    case Logger = "logger"
    case SessionDelegate = "sessionDelegate"
    case Path = "path"
    case ExtraHeaders = "extraHeaders"
    case HandleQueue = "handleQueue"
}
```

Would need a convenience init on the client for Objective-C

lightsprint09:
Like this?

```swift
enum SocketInitialOptionsEnum {
    case ConnectionParameter([String: AnyObject]?)
    case Reconnects(Bool)

    func toString() -> String {
        switch self {
        case ConnectionParameter:
            return "connectParams"
        case Reconnects:
            return "reconnects"
        }
    }
}
```

nuclearace:
I was thinking something more along the lines of https://github.com/socketio/socket.io-client-swift/commit/40ac53844d908e65cfb870ca27cfa2356c3034e1

nuclearace:
That way also has the benefit in that it won't be a compatibility breaking change.

SocketIOClientSwift/SocketInitialOptions.swift
```diff
@@ -0,0 +1,26 @@
...
+//  InitialOptions.swift
+//  Socket.IO-Client-Swift
+//
+//  Created by Lukas Schmidt on 17.10.15.
+//
+//
+
+import Foundation
+
+struct SocketInitialOptions {
```
NachoSoto: If you make this an `enum` with associated `String` values instead, you can statically guarantee that all keys are unique.

nuclearace:
#212 Is probably the better way to go about it.

PR #212, opened 2015-10-17
Better options

nuclearace:
- [x] Update readme


lightsprint09:
this solution is defiantly better then mine.

But it is not Typesave. Wait a few days. I want to explore some more typesave solutions

nuclearace:
@lightsprint09 What do you think?

SocketIOClientSwift/SocketIOClient.swift
```diff
@@ -68,45 +68,41 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
...
+                reconnectWait = wait
+            case .Nsp(let nsp):
+                self.nsp = nsp
+            case .Log(let log):
+                Logger.log = log
+            case .Logger(let logger):
+                Logger = logger
+            case .HandleQueue(let queue):
+                handleQueue = queue
+            default:
```
lightsprint09: this shouldn't be necessary

lightsprint09:
I like your solution.

I don't like nsp for a shortcut. It doesn't seem clear to me

SocketIOClientSwift/SocketIOClientOption.swift
```diff
@@ -0,0 +1,123 @@
...
+            return .ExtraHeaders(value as! [String: String])
+        case "handleQueue" where value is dispatch_queue_t:
+            return .HandleQueue(value as! dispatch_queue_t)
+        default:
+            return nil
+        }
+    }
+    
+    static func getSocketIOOptionValue(option: SocketIOClientOption) -> AnyObject? {
+        if let value = Mirror(reflecting: option).children.first?.value as? AnyObject {
```
lightsprint09: I don`t understand what this line does.

SocketIOClientSwift/SocketIOClient.swift
nuclearace: It's needed because in Swift switches must be exhaustive. And because we don't care about every option possible here, we need a default.

lightsprint09: oh I see. Some options are only used by the engine?

SocketIOClientSwift/SocketIOClientOption.swift
nuclearace: It's some black magic reflection to the value regardless of the case.

SocketIOClientSwift/SocketIOClient.swift
nuclearace: Yes

SocketIOClientSwift/SocketIOClientOption.swift
lightsprint09: ok. 
```swift
return Mirror(reflecting: option).children.first?.value as? AnyObject
```
would this work?

nuclearace: Yeah

SocketIOClientSwift/SocketIOClientOption.swift
```diff
@@ -0,0 +1,119 @@
...
+        
+        for option in set {
+            options[option.description] = getSocketIOOptionValue(option)
+        }
+        
+        return options
+    }
+}
+
+public func ==(lhs: SocketIOClientOption, rhs: SocketIOClientOption) -> Bool {
```
lightsprint09: Is this necessary anymore?

SocketIOClientSwift/SocketIOClient.swift
```diff
@@ -52,12 +50,14 @@ public final class SocketIOClient: NSObject, SocketEngineClient {
...
+    private(set) var handleQueue = dispatch_get_main_queue()
+    private(set) var reconnectAttempts = -1
     var waitingData = [SocketPacket]()
     
     /**
-    Create a new SocketIOClient. opts can be omitted
+    Type safe way to create a new SocketIOClient. opts can be omitted
     */
-    public init(var socketURL: String, opts: [String: AnyObject]? = nil) {
+    public init(var socketURL: String, opts: SocketOptionsSet? = nil) {
```
lightsprint09: i would rename it to options...
-> I don't like shortcuts

SocketIOClientSwift/SocketTypes.swift
```diff
@@ -27,8 +27,9 @@ import Foundation
 public typealias AckCallback = ([AnyObject]) -> Void
 public typealias NormalCallback = ([AnyObject], SocketAckEmitter?) -> Void
 public typealias OnAckCallback = (timeoutAfter: UInt64, callback: AckCallback) -> Void
+public typealias SocketOptionsSet = Set<SocketIOClientOption>
```
lightsprint09: I this a clear solution.
Because I don't see at first sight that this should be a Set.

SocketIOClientSwift/SocketIOClientOption.swift
nuclearace: It's required by hashable

SocketIOClientSwift/SocketIOClient.swift
nuclearace: Yeah

SocketIOClientSwift/SocketIOClientOption.swift
lightsprint09: ok.

SocketIOClientSwift/SocketTypes.swift
nuclearace: Opposed to what? I don't see a case where you'd have multiple items for the same option.

lightsprint09: I would remove the typealias and use ```Set<SocketIOClientOption>``` directly in the API because everybody knows about Set but nobody knows about this secret SocketOptionsSet

nuclearace: Valid point

nuclearace:
Also I guess travis is dead

lightsprint09:
It is... https://www.traviscistatus.com

nuclearace:
This good for merge?

lightsprint09:
I think so

nuclearace:
It's going to be a major version change I think, since we changed the constructor interface

lightsprint09:
Mh yeah. Now is would be a good time to ship other API changes...

PR #213, opened 2015-10-18
Real type save

lightsprint09:
This would be a real typesave solution. It is a prove of concept. It works but I haven´t implement all option keys. This won't break normal API and add a more convenient API on top.

I would like to hear your feedback


nuclearace:
Yeah, I like your idea. The only concern I have is it's going to throw a lot of people off, since they're probably used to doing it the old way.

lightsprint09:
But you can use the old API as well as the new one

nuclearace:
Hm, I'll toy around with this approach a bit in the other branch

PR #248, opened 2015-12-02
Engine refactoring

nuclearace:


SocketIOClientSwift/SocketEngine.swift
```diff
@@ -33,6 +33,8 @@ public final class SocketEngine: NSObject, SocketEngineSpec, WebSocketDelegate {
     public private(set) var ws: WebSocket?
 
     public weak var client: SocketEngineClient?
+    
+    private weak var sessionDelegate: NSURLSessionDelegate?
```
NachoSoto: Why is this weak? the delegate won't have a strong reference to the engine so there's no risk of retain cycle.

SocketIOClientSwift/SocketIOClientOption.swift
```diff
@@ -68,6 +69,8 @@ public enum SocketIOClientOption: ClientOption {
             return .ReconnectAttempts(value as! Int)
         case "reconnectWait" where value is Int:
             return .ReconnectWait(value as! Int)
+        case "forceNew" where value is Bool:
```
NachoSoto: I'd recommend doing `switch (key, value)`, because then you can do `case ("forceNew", let value) where value is Bool)` and will automatically have the right type.

SocketIOClientSwift/SocketEngine.swift
nuclearace: The NSURLSessionDelegate could also be the SocketEngineClient

nuclearace: If someone wanted to build off the engine.

SocketIOClientSwift/SocketIOClientOption.swift
nuclearace: Are you sure that's valid? 

```swift
case ("connectParams", let params) where params is [String: AnyObject]:
    return .ConnectParams(params) // still wants to cast
```

nuclearace: Also this option really doesn't make sense anymore. Maybe I'm not understanding how socket.io-client is supposed to handle reconnects (the protocol is kinda sketchy,) but it seems that it should always get a new sid on reconnect. So this option would only be good if there's some bug in the engine after a reconnect attempt, and you want to have a new engine every time.

PR #262, opened 2015-12-11
Avoid doing unaligned reads/writes via UnsafePointer

mstorsjo:
Instead read/write individual bytes via UnsafePointer<UInt8>.

This fixes one case of crashes that only happen when optimization is turned on. (Possibly related to issue #240 and others?)

Not all processors support unaligned reads/writes of 64 bit values. Without optimizations, the reads/writes probably end up touching one byte at a time (like this commit makes it do), while higher optimization levels convert it to single reads/writes of the full size, which may require that the pointer is properly aligned.

### DIFF
--- a/Source/WebSocket.swift
+++ b/Source/WebSocket.swift
@@ -421,6 +421,33 @@ public class WebSocket : NSObject, NSStreamDelegate {
         return false
     }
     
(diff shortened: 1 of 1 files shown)


nuclearace:
This should be merged into https://github.com/daltoniam/Starscream first, then I will update

PR #265, opened 2015-12-22
Change product names to "Socket_IO_Client_Swift" to work with Carthage

sotownsend:
When using this framework as a dependency in Cocoa-Pods, the frameworks product name is changed from "SocketIO" to "Socket_IO_Client_Swift".  If you then have a carthage supported framework, that requires this as a dependency, then the names will be different when using carthage as a dependency manager.

### DIFF
--- a/Socket.IO-Client-Swift.xcodeproj/project.pbxproj
+++ b/Socket.IO-Client-Swift.xcodeproj/project.pbxproj
@@ -8,7 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		572EF21F1B51F16C00EEBB58 /* SocketIO-iOS.h in Headers */ = {isa = PBXBuildFile; fileRef = 572EF21E1B51F16C00EEBB58 /* SocketIO-iOS.h */; settings = {ATTRIBUTES = (Public, ); }; };
-		572EF2251B51F16C00EEBB58 /* SocketIO.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 572EF2191B51F16C00EEBB58 /* SocketIO.framework */; };
+		572EF2251B51F16C00EEBB58 /* Socket_IO_Client_Swift.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 572EF2191B51F16C00EEBB58 /* Socket_IO_Client_Swift.framework */; };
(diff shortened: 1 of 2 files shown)


NachoSoto:
I use this With Carthage and the name has not been a problem. Why would you change the name?

sotownsend:
Hey @NachoSoto 

Yes, you are correct.  I have not had issues with issuing this directly with carthage.  Let me illustrate when I do have an issue...

If I have a custom framework, hosted on Github, that uses CocoaPods to construct it's dependencies:
screen shot 2015-12-22 at 5 03 16 am

The resulting framework, when built by the xcodeproject, has a dynamic import for the framework at the path `@rpath/Socket_IO_Client_Swift.Socket_IO_Client_Swift.framework`

Now if this framework is then taken and used in a carthage project, I have to tell the user to include the socket-io swift framework:

screen shot 2015-12-22 at 5 03 21 am

However, when carthage imports this time, the framework names do not match because carthage builds the framework with the product name as `SocketIO`.

nuclearace:
Wouldn't the other targets also need to be changed

NachoSoto:
That seems like a problem with CocoaPods, who's not respecting the product name.

nuclearace:
I just tested the Swift Package Manager for this project and I can say that it was much easier to use than Carthage or CocoaPods was. That being said, it'll be awhile before it's to the point where it could replace Carthage or CocoaPods.

sotownsend:
@NachoSoto I agree with your sentiment, but that is a question of philosophy :)

Could you please reconsider this pull request?  Pretty please?  :)  Alternatively, we could add a flag to the pod-spec to change the product-name to SocketIO.

NachoSoto:
It's not a philosophy problem, I'm being 100% serious: if `CocoaPods` isn't respecting `Xcode`'s `product name` setting then that's something that needs to be either fixed in `CocoaPods`, or make sure that the `Podfile` is correct, not masking the problem by changing the product name to whatever `CocoaPods` happens to have right now.

nuclearace:
Honestly all the framework builds should be renamed so that they become `SocketIOClientSwift`

NachoSoto:
I don't think it makes sense to suffix a framework with the word "Swift". It's irrelevant to a client of the module what language was used to compile it.

nuclearace:
The only reason is because that's the actual name of the project.

nuclearace:
Will be fixed in #278

PR #277, opened 2016-01-14
add new method engineDidOpen in socketEngineClient protocol

leo0307vb:
Add new method to socketEngineClient to use engine.io without the socket.io


nuclearace:
Seems a little odd to only put the method call inside the websocket method. It should probably be in SocketEngine.handleOpen somewhere. Since by default socket.io should start by using long-polling and then upgrade to WebSocket

nuclearace:
Also the method should not be optional. And yet do !

leo0307vb:
I updated the call to engineDidOpen and set it not optional

nuclearace:
I think for PR the implementation that the client gives should be empty. Since there's actually another packet that says the socket.io connection is connected

PR #278, opened 2016-01-14
Standardize module name

nuclearace:


nuclearace:
@banjun If you're still around, is 3587c20 the right way to change the name of the framework CocoaPods spits out?

banjun:
@nuclearace As `name` is the key for your pod, use `module_name` to change its framework name.
see for example:
https://github.com/banjun/ikemen/blob/1765b23b900dcf279b4263361bb4026e8605a0bb/※ikemen.podspec

nuclearace:
@banjun thanks

PR #317, opened 2016-02-26
fix timeoutAfter handler always close connection

jtjacka:
Currently the status check always results in true therefore the socket is closed every-time a timeout is evaluated


nuclearace:
Oops, thanks

PR #343, opened 2016-04-10
Websocket compression [DONT MERGE]

objectiveSee:
This Pull Request adds support for websocket compression, which was added to Socket.io in version 1.4. This is a basic implementation which still has some outstanding issues, therefore `[DONT MERGE]` was added to the title of the issue.

-  Basic support for `Sec-WebSocket-Extensions` header and the negotiation of the `permessage-deflate` extension.
- Websocket compression using the `zlib` library. Required using an Objective-C helper class to interface with `zlib` as it wasn't possible or easy to use that library's C API from within Swift.

Remaining issues:
- [ ] Use https://github.com/tidwall/DeflateSwift or similar instead of ObjC
- [ ] Turn compression off by default and expose API to enable it.
- [ ] Verify that `response.bytesLeft -= Int(dataLength)` is correct for handling partial frames. Test when extra data is present (ie. next frame)
- [ ] Pick error code to use when compressed frames are unable to be decompressed.
- [ ] Handle error cases and TODOs in the DeflateHelper class. There are also optimizations that can be done such as not building a new inflator with each packet.
- [ ] Determine whether we should re-use the deflate context to achieve better compression. From Because the client and server may reuse their DEFLATE context between messages, the compression improves as more similar messages are sent over the connection. See https://blog.jcoglan.com/2015/03/30/websocket-extensions-as-plugins/ (`Because the client and server may reuse their DEFLATE context between messages...`)
- [ ] Requires more testing.
- [ ] Change version in Podspec

### DIFF
--- a/Socket.IO-Client-Swift.podspec
+++ b/Socket.IO-Client-Swift.podspec
@@ -15,7 +15,8 @@ Pod::Spec.new do |s|
   s.osx.deployment_target = '10.10'
   s.tvos.deployment_target = '9.0'
   s.source       = { :git => "https://github.com/socketio/socket.io-client-swift.git", :tag => 'v6.0.0' }
-  s.source_files  = "Source/**/*.swift"
+  s.source_files  = "Source/*"
   s.requires_arc = true
   # s.dependency 'Starscream', '~> 0.9' # currently this repo includes Starscream swift files
+  s.library = 'z'
 end
--- a/Socket.IO-Client-Swift.xcodeproj/project.pbxproj
+++ b/Socket.IO-Client-Swift.xcodeproj/project.pbxproj
@@ -7,6 +7,9 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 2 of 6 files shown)


nuclearace:
I'm not accepting anything that's in Objective-C. Should be rewritten in Swift. Also, this would need to go added to the upstream WebSocket, and then I'd update the dependency.

nuclearace:
https://github.com/daltoniam/Starscream/issues/169

objectiveSee:
@nuclearace I'd be happy to port changes into starscream, though it would be better to just include that as a sub-pod. What's the current reason for not using starscream as a dependency on this pod?

nuclearace:
Because it makes it more difficult to support other package managers and installing manually more difficult. And also because I want to support swift3 ahead of time. https://github.com/socketio/socket.io-client-swift/blob/swift3/Source/WebSocket.swift and it makes it easier to make changes releated to that.

yosiat:
@nuclearace Hi, do you still open for pull requests to add per message deflate support?

PR #349, opened 2016-04-14
How to know how many events i have registered? 

carlosen14:
Excellent tool to work with, except i have a request.. 

In the android development they have `socket.listeners();` and this gives you a `List<Listeners>`..

would it be possible of doing the same whit the Swift version? 
actually what i am doing is this..

`
extension SocketIOClient
{
  func isConnected() -> Bool {
    if ( self.status  != SocketIOClientStatus.Connected ){
      return false;
    }else{
      return true
    }
  }
  
  func removeHandler(nameEvent: String) -> Void {
    if (self.isConnected()){
      self.emitWithAck(nameEvent, [])(timeoutAfter: 5) {data in
        self.removeAllHandlers()
        let main = MainScreenVC()
        main.addListeners(except: nameEvent)
      }
    }
  }
}
`

Hope you can help!

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ Carthage
 -----------------
 Add this line to your `Cartfile`:
 ```
-github "socketio/socket.io-client-swift" ~> 6.0.0 # Or latest version
+github "socketio/socket.io-client-swift" ~> 6.1.0 # Or latest version
 ```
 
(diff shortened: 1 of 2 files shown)


nuclearace:
Why is this a pull request, it should be an Issue.

PR #361, opened 2016-05-04
Fixing url encoding for connect params

yannickl:
Hi,

I just updated my library from v4.x to v6.x and you re-introduce an issue with the connect params url encoding.

So here is my patch.

Best


nuclearace:
What kind of connect params cause it to mess up? Because it looks to me that the query is being escaped

nuclearace:
Nvm I see.

yannickl:
Sorry I was out, I just added a test case to check that.

The problem can occur when you pass a serialised date for example:
```swift
[
    "created": "2016-05-04T18:31:15+0200"
]
```

Thank you for the merge. 👍

PR #372, opened 2016-05-23
Fix crash if error contains formatting strings.

ATLCTO:


nuclearace:
When does this crash?

ATLCTO:
Our server went down and it crashed the iOS app because % in 100% is a formatting string.

"Got unknown error from server <!DOCTYPE html>\n    <html>\n    <head>\n      <meta name=\"viewport\" content=\"width=device-width, initial-scale=1\">\n      <style type=\"text/css\">\n        html, body, iframe { margin: 0; padding: 0; height: 100%; }\n        iframe { display: block; width: 100%; border: none; }\n      </style>\n    <title>Application Error</title>\n    </head>\n    <body>\n      <iframe src=\"//s3.amazonaws.com/heroku_pages/error.html\">\n        <p>Application Error</p>\n      </iframe>\n    </body>\n    </html>"

nuclearace:
Oh I see

nuclearace:
Actually I'm having trouble reproducing. Just calling didError with that string doesn't seem to crash mine.

ATLCTO:
image

PR #380, opened 2016-05-31
Add Proxy support for websocket

oldshuren:
This pull request add proxy support for socket.io client behind a proxy server.

The proxy configuration is getting from the system. The proxy server must support the CONNECT method.

When determined that there is proxy setting, instead of opening a network stream to the remote websocket server, it opens a network stream to the proxy server, then issuing an HTTP CONNECT request. If the proxy server responds with 200 OK, it treats this network stream as a normal stream to the remote websocket server.

### DIFF
--- a/Socket.IO-Client-Swift.podspec
+++ b/Socket.IO-Client-Swift.podspec
@@ -14,7 +14,7 @@ Pod::Spec.new do |s|
   s.ios.deployment_target = '8.0'
   s.osx.deployment_target = '10.10'
   s.tvos.deployment_target = '9.0'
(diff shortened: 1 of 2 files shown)


nuclearace:
All pull request for WebSocket should be made at https://github.com/daltoniam/Starscream. Also you shouldn't change podspec to point to yours, I would create a new version and push it.

oldshuren:
Ok, thanks!

I modified podspec because 'path' and 'branch' attribute in my Podfile
didn't work. But even I changed the podspec, it still didn't work. I'll
changed it back.

neiker:
Here's the PR on Starscream repo:
https://github.com/daltoniam/Starscream/pull/547

PR #397, opened 2016-06-16
Updated code to support Swift 2.3

davbeck:
There were some issues with the tests when I went to create the migrations. Some of the targets included the libraries actual source files in the test targets, but also imported the module. That has been resolved and all tests are passing.

I'm submitting this request to master, but I believe that there should be a 2.3 branch instead. What I would propose as the timeline would be:

1. Maintain 2.3 and 3.0 branches until Xcode 8 GM is released.
2. Keep the 2.3 branch long term for legacy support.
3. Merge 3.0 into master.
4. When the next version of Swift is released, create a legacy swift-3.0 branch off of master.

This way, user's of the framework can target a specific branch if they need legacy support, or master if they want the most current version.

Thoughts?

### DIFF
--- a/Socket.IO-Client-Swift.xcodeproj/project.pbxproj
+++ b/Socket.IO-Client-Swift.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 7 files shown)


nuclearace:
Can you open this against the 2.3 branch I just made?

PR #422, opened 2016-07-13
Enable APPLICATION_EXTENSION_API_ONLY

aputinski:
Trying to use this in an iMessage app and ran into the `linking against a dylib which is not safe for use in application extensions` warning.


nuclearace:
Is it just a warning or does it cause the build to fail?

aputinski:
It's just a warning, but the app will get rejected from the AppStore

aputinski:
@nuclearace any thoughts on merging this?

PR #427, opened 2016-07-20
Xcode 8 Beta 3

aputinski:
Taking a look at https://github.com/ReactiveX/RxSwift/commit/824f23c94e34577d2259c40a5575f509e24e50cb, it seems that setting `SWIFT_VERSION` is required to compile with Xcode 8 Beta 3

### DIFF
--- a/Source/SocketEngine.swift
+++ b/Source/SocketEngine.swift
@@ -151,8 +151,7 @@ public final class SocketEngine : NSObject, SocketEnginePollable, SocketEngineWe
         if message.hasPrefix("b4") {
             // binary in base64 string
             let noPrefix = message[message.characters.index(message.startIndex, offsetBy: 2)..<message.endIndex]
-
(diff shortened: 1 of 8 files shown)


Source/SocketEngine.swift
```diff
@@ -151,8 +151,7 @@ public final class SocketEngine : NSObject, SocketEnginePollable, SocketEngineWe
         if message.hasPrefix("b4") {
             // binary in base64 string
             let noPrefix = message[message.characters.index(message.startIndex, offsetBy: 2)..<message.endIndex]
-
-            if let data = Data(base64Encoded: noPrefix, options: Data.Base64EncodingOptions(rawValue: 0)) {
+            if let data = Data(base64Encoded: noPrefix, options: Data.Base64DecodingOptions(rawValue: 0)) {
```
aputinski: This was the only actual error. Does this seem correct? `Data.Base64EncodingOptions` => `Data.Base64DecodingOptions`

PR #436, opened 2016-07-25
Development

drekka:
I have an Objective-C framework project which uses socket.io. I then have an Objective-C app project that uses the framework project (but not socket.io directly). Both projects use carthage to handle dependencies. 

When I do a carthage update of the framework project,everything compiles fine.

When I do a carthage update of the app project, the compile of socket.io fails because it is unable to put a SecKeyRef struct into an NSArray. 

This pull request makes a method that returns a `[Seckeyref]` and a property of type `[SecKeyRef]` private so that they are removed from the generated Objective-C header file. So far this seems to have fixed my compilation problem.

### DIFF
--- a/Source/SSLSecurity.swift
+++ b/Source/SSLSecurity.swift
@@ -241,7 +241,7 @@ public class SSLSecurity : NSObject {
      
      - returns: the public keys from the certifcate chain for the trust
      */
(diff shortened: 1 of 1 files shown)


nuclearace:
Mark them as @nonobjc, not private

drekka:
Thanks

PR #443, opened 2016-08-05
Now working with Carthage and XCode 8 beta 4

drekka:
Had to make a few changes to get it working again after getting errors based on not being able to find the umbrella header.
 * Ran the covert to Swift 2.3.
 * Set support legacy swift version in build settings.
 * Make the iOS, tvOS and Mac targets all build frameworks with the same name so that the shared unit tests can all run. Previously they would fail to compile depending on what framework name was being imported.
 * Fixed missing info.plist files in tests.
 * Fixes some code which was warning about setting an enum with an int. Set to the default (0) Invalid value of the enum.
 * All tests working for all targets.
 * Imported main Swift header in umbrella header file. 
 * Removed socket.io source files from unit test target compiles.

### DIFF
--- a/Socket.IO-Client-Swift.xcodeproj/project.pbxproj
+++ b/Socket.IO-Client-Swift.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 46;
+	objectVersion = 48;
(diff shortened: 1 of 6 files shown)


drekka:
Dam, still not working in final app. Libswiftcore not found. Closing until I've resolved this.

PR #454, opened 2016-08-14
swift compile time optimizations

eventomer:
reducing about 20 seconds of compilation time because of swift type checker string concatenation issue

### DIFF
--- a/Source/SocketEngine.swift
+++ b/Source/SocketEngine.swift
@@ -243,8 +243,8 @@ public final class SocketEngine : NSObject, NSURLSessionDelegate, SocketEnginePo
             }
         }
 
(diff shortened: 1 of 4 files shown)


nuclearace:
What, I've never seen it take more than 7 or 8 seconds to compile.

nuclearace:
Also, you'll want to branch off of development and make changes there.

MeGaPk:
@nuclearace you are lucky... I have 20-30 seconds... on my macbook pro.

nuclearace:
Going to close this, since compile time should be a Swift issue itself.

hfossli:
If this helps reducing 20 seconds of compilation now and the code is not suffering readability then I don't understand why it can't be merged. When is the "free string compilation time lunch" coming?

PR #466, opened 2016-08-23
Switch to using Any instead of AnyObject

davbeck:
In Swift 3 AnyObject is rarely what you want to use. Objective-C API gets imported as Any and while types like Array and Dictionary can be coerced to AnyObject, they prefer Any, which doesn’t require a manual cast.


nuclearace:
Thanks, that makes things look less suck

PR #489, opened 2016-09-11
Propagate websocket connected event to SocketIOClientStatus when forced websocket connection

therudite:


nuclearace:
Shouldn't the open packet be sent over WebSocket anyway?

therudite:
Yes, it is sending and websocket is connected as well but this is not getting reflected in socket status.

nuclearace:
Right, but is a connect event being emitted when forcing websocket? Because the websocket methods shouldn't be sending events to the client directly, they should come through the websocket and be sent the handle method.

nuclearace:
https://github.com/socketio/socket.io-client-swift/blob/development/Source/SocketIOClient.swift#L188

https://github.com/socketio/socket.io-client-swift/blob/development/Source/SocketParsable.swift#L39

Issue #505, opened 2016-09-20
8.0.1 crash


brightsider:
<img width="1423" alt="crash" src="https://cloud.githubusercontent.com/assets/4996231/18663280/810068ba-7f26-11e6-8b92-5a27f06deaaf.png">

Gagster:
Hi, Where can I get the version 8.0.1?
I've tried to use pod, but it saids that the latest one is 8.0.0 
: (  plz Help~

PR #515, opened 2016-09-27
Fix crash

brightsider:
It's a fix for #505.
For 8.0.2 this crash reproduced too.


nuclearace:
Can you open this up against https://github.com/daltoniam/Starscream so it can be reviewed there?

PR #523, opened 2016-09-29
Fixed parameter names in documentation

ujell:


nuclearace:
This would be better opened on https://github.com/daltoniam/Starscream

PR #557, opened 2016-10-28
add SocketAckEmitter.expected to indicate if an ack is expected by server

naderio:
this is needed to send acknowledgement when required and to make NativeScript implementation mimic JavaScript's https://github.com/naderio/nativescript-socket.io/blob/64630685f2a000c79e116ed9e07662638bc32456/socket.ios.ts#L82

### DIFF
--- a/Source/SocketAckEmitter.swift
+++ b/Source/SocketAckEmitter.swift
@@ -44,6 +44,10 @@ public final class SocketAckEmitter : NSObject {
         
         socket.emitAck(ackNum, with: items)
     }
+        
+    public func isRequired() -> Bool {
+        return ackNum != -1;
+    }
 }
 
 public final class OnAckCallback : NSObject {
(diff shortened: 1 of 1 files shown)


Source/SocketAckEmitter.swift
```diff
@@ -44,6 +44,10 @@ public final class SocketAckEmitter : NSObject {
         
         socket.emitAck(ackNum, with: items)
     }
+        
+    public func isRequired() -> Bool {
+        return ackNum != -1;
```
nuclearace: semicolon's are bad

nuclearace: This should probably be a computed property rather than a method.

```swift
var isRequired: Bool {
    return ackNum != -1
}
```

But to take a step back. I don't remember required acks being a thing. Were they recently added?

naderio: in the JavaScript API, when an event expecting an ack is sent by server, the client provides the ack callback as last argument in event handler.
not a way to "require" an ack .. more like "an ack is expected"

nuclearace:
One last thing, this should opened against development rather than master.

Issue #89, opened 2015-05-29
Memory Leaks


packedbox:
Hi, i'm using the swift library in an ObjectiveC App to stream data from the phone to a socket, so i emit quite frequently an I see my the memory used by the app growing linearly. 

``` javascript
@property SocketIOClient *socket;
```

``` javascript
[_socket emit:@"GSR" withItems:@[[NSNumber numberWithDouble:timestamp],[NSNumber numberWithFloat:gsr]]];
```

![capture d ecran 2015-05-29 a 12 55 53](https://cloud.githubusercontent.com/assets/1764662/7881331/1a65e934-0602-11e5-99e8-92660e8d1348.png)

If i comment out the write call in the send function in SocketEngine.swift, the memory is stable

``` javascript
   /// Send an engine message (4)
    public func send(msg:String, withData datas:ContiguousArray<NSData>?) {
        if probing {
            probeWait.append((msg, PacketType.MESSAGE, datas))
        } else {
           // write(msg, withType: PacketType.MESSAGE, withData: datas)
        }
    }
```

![capture d ecran 2015-05-29 a 16 49 10](https://cloud.githubusercontent.com/assets/1764662/7885170/ab1a256a-0622-11e5-9ea9-5c99376389f0.png)

nuclearace:
Commenting out that line should actually break things, since it's not sending any messages. But I guess it shows that it's happening somewhere during sending.  Does this happen when using polling or WebSockets?

packedbox:
Yes you’re right it was just to check were the leaks could come from.
I tried to go deeper in the code to check what could be wrong but I must admit i’m a complete newbie in Swift and the code is using async queues so It’s not easy to debug. 

The leaking behavior is present with all combinaisons of forcepolling and forcewebsocket. I tried to use the instruments to track the leak 

setting only forcepolling gives a slighly different pattern in the memory view ( no "steps" in the memory )
![capture d ecran 2015-05-29 a 17 14 40](https://cloud.githubusercontent.com/assets/1764662/7885997/424966a4-0627-11e5-82e2-b60e250b4a66.png)

nuclearace:
How often are you emitting?

packedbox:
I stream data that I receive from a BLE device the rate should be around 60hz

nuclearace:
I think 60 times a second is kinda insane for socket.io. Have you tried doing any instruments profile?

nuclearace:
Also is this the latest version of the client?

[...]

PR #620, opened 2017-02-22
synchronized call to mutating ackHandlers.executeAck

alevinru:
helps with memory issues of socketio/socket.io-client-swift#89

There's no need to executeAck async because it is async internally, but dangerous because it's mutating


nuclearace:
Needs to be opened on development

PR #629, opened 2017-03-06
SocketAckManager thread-safe acks removal

alevinru:
One more semaphore helps to prevent from calling a fulfilled ack's callback from timeoutAck() in harsh multi-threaded conditions.

We concurrently do about 1500 emits within about 20 seconds each with a timeout of 30 seconds. After all the emits are successfully called back, sometimes there are some items still remain in the SocketManager private acks Set. After that the time comes to firing the timeoutAck () and then it finds a junk acks item and tries to call it back. But at the moment the callback's closure is already deallocated and there's a crash.

### DIFF
--- a/Source/SocketAckManager.swift
+++ b/Source/SocketAckManager.swift
@@ -51,6 +51,7 @@ private func ==(lhs: SocketAck, rhs: SocketAck) -> Bool {
 
 struct SocketAckManager {
     private var acks = Set<SocketAck>(minimumCapacity: 1)
+    private let ackSemaphore = DispatchSemaphore(value: 1)
     
     mutating func addAck(_ ack: Int, callback: @escaping AckCallback) {
         acks.insert(SocketAck(ack: ack, callback: callback))
@@ -58,6 +59,8 @@ struct SocketAckManager {
     
     /// Should be called on handle queue
     mutating func executeAck(_ ack: Int, with items: [Any], onQueue: DispatchQueue) {
+        ackSemaphore.wait()
+        defer { ackSemaphore.signal() }
         let ack = acks.remove(SocketAck(ack: ack))
         
         onQueue.async() { ack?.callback(items) }
@@ -65,8 +68,12 @@ struct SocketAckManager {
     
     /// Should be called on handle queue
     mutating func timeoutAck(_ ack: Int, onQueue: DispatchQueue) {
+        ackSemaphore.wait()
+        defer { ackSemaphore.signal() }
         let ack = acks.remove(SocketAck(ack: ack))
         
-        onQueue.async() { ack?.callback?(["NO ACK"]) }
+        onQueue.async() {
+            ack?.callback?(["NO ACK"])
+        }
     }
 }


Source/SocketAckManager.swift
```diff
@@ -51,22 +51,29 @@ private func ==(lhs: SocketAck, rhs: SocketAck) -> Bool {
...
+    private let ackSemaphore = DispatchSemaphore(value: 1)
     
     mutating func addAck(_ ack: Int, callback: @escaping AckCallback) {
         acks.insert(SocketAck(ack: ack, callback: callback))
     }
     
     /// Should be called on handle queue
     mutating func executeAck(_ ack: Int, with items: [Any], onQueue: DispatchQueue) {
+        ackSemaphore.wait()
+        defer { ackSemaphore.signal() }
```
nuclearace: Can we just put the signal right after the remove?

Source/SocketAckManager.swift
```diff
@@ -51,22 +51,29 @@ private func ==(lhs: SocketAck, rhs: SocketAck) -> Bool {
...
         let ack = acks.remove(SocketAck(ack: ack))
         
         onQueue.async() { ack?.callback(items) }
     }
     
     /// Should be called on handle queue
     mutating func timeoutAck(_ ack: Int, onQueue: DispatchQueue) {
+        ackSemaphore.wait()
+        defer { ackSemaphore.signal() }
```
nuclearace: Can we put the signal right after the remove?

alevinru: Yes, we can, but the order doesn't matter with defer.  Defer makes it signalling after the func execution's over.

alevinru: I suppose it's important to use defer here because as far as i know mutating funcs actually mutate after the function has finished.

nuclearace: Mutate what? If you can find some source for that. Because that would be incredibly surprising. I'm thinking about rewriting the queue system anyway, there are way too many of them.

alevinru: https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/Methods.html

Chapter "Modifying Value Types from Within Instance Methods"

> The method can then mutate (that is, change) its properties from within the method, and any changes that it makes are written back to the original structure when the method ends.

nuclearace: Thanks. Yeah that makes me want to redo the queue structure for the entire client even more. Rather than having like 3 or 4 queues per entity, just have 1 or 2.

nuclearace: But this looks good until that is done.

alevinru: For the queueing there are nice built-in classes OperationQueue and Operation. Apple says it's thread-safe and it seems so with my experience. Just need to subclass and implement KVO for the properties and a pair of methods as it's written in docs.

nuclearace: I'd rather avoid KVO. I'm not a fan of it.

alevinru: KVO is ugly indeed, but it's only a dozen of lines about it when subclassing an asynchronous Operation with no obligation to read these lines after. This is the cost of OperationQueue benefits. It's like a kind of oblation :)

Issue #630, opened 2017-03-07
Received the same event twice from one socketClient


musicabbage:
Hi, I received the same event twice from one socketClient object after receonnected... it usually happens on bad networking.
I can't reproduce the issue in a regular pattern, but in my guess it's caused by receiving data from old websocket/socketEngine object(maybe leak or not be released yet) then it will trigger socketClient's event also, so socketClient receive event twice but one of which is useless and should be close.

Is there anyway to check if the event is triggered by the websocket/socketEngine owned with SocketClient?

musicabbage:
My config params

`    NSDictionary *options = @{@"forceWebsockets":@YES
                              ,@"reconnects":@YES
                              ,@"reconnectWait":@1
                              ,@"forceNew":@YES
                              };`

lishuanglong521:
How to solve the problem?

musicabbage:
clear up ur problem first, did u have this issue in a regular way?
If yes, make sure u didn't added listeners multi times as @terenceLuffy said.
If not... I can help u nothing bcoz I'm trying to figure it out as well 😂😂😂😂

PR #651, opened 2017-04-13
remove engine’s listener before assign new instance

musicabbage:
#630

### DIFF
--- a/Source/SocketEngine.swift
+++ b/Source/SocketEngine.swift
@@ -244,10 +244,7 @@ public final class SocketEngine : NSObject, URLSessionDelegate, SocketEnginePoll
 
     private func createWebsocketAndConnect() {
         
-        if let oldWs = ws {
-            oldWs.delegate = nil
-        }
-        
+        ws?.delegate = nil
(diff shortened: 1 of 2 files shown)


Source/SocketEngine.swift
```diff
@@ -243,6 +243,11 @@ public final class SocketEngine : NSObject, URLSessionDelegate, SocketEnginePoll
     }
 
     private func createWebsocketAndConnect() {
+        
+        if let oldWs = ws {
```
nuclearace:
```swift
ws?.delegate = nil
```

Source/SocketIOClient.swift
```diff
@@ -120,6 +120,10 @@ public final class SocketIOClient : NSObject, SocketEngineClient, SocketParsable
     private func addEngine() -> SocketEngineSpec {
         DefaultSocketLogger.Logger.log("Adding engine", type: logType, args: "")
 
+        if let oldEngine = engine {
```
nuclearace:
```swift
engine?.client = nil
```

PR #660, opened 2017-04-26
add listeners for pingSent & pongReceived

faizan-carameltech:
I needed to make some decisions on basis of number of failed pings and pong timeouts. Therefore publicly handling listeners for these events.

### DIFF
--- a/Source/SocketEngineClient.swift
+++ b/Source/SocketEngineClient.swift
@@ -26,8 +26,8 @@
 import Foundation
 
 @objc public protocol SocketEngineClient {    
-    func engineDidSentPing()
-    func engineDidReceivedPong()
+    func engineDidSendPing()
+    func engineDidReceivePong()
     func engineDidError(reason: String)
     func engineDidClose(reason: String)
     func engineDidOpen(reason: String)
--- a/Source/SocketEngineClient.swift
+++ b/Source/SocketEngineClient.swift
@@ -26,6 +26,8 @@
 import Foundation
 
 @objc public protocol SocketEngineClient {    
+    func engineDidSentPing()
+    func engineDidReceivedPong()
     func engineDidError(reason: String)
     func engineDidClose(reason: String)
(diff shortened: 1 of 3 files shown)


Source/SocketEngineClient.swift
```diff
@@ -26,6 +26,8 @@
 import Foundation
 
 @objc public protocol SocketEngineClient {    
+    func engineDidSentPing()
```
nuclearace: `engineDidSendPing`

Source/SocketEngineClient.swift
```diff
@@ -26,6 +26,8 @@
...
+    func engineDidReceivedPong()
```
nuclearace: `engineDidReceivePong`

Source/SocketIOClient.swift
```diff
@@ -275,6 +275,19 @@ public final class SocketIOClient : NSObject, SocketEngineClient, SocketParsable
         }
     }
 
+    ///ping & pong
+    public func engineDidSentPing() {
+        DefaultSocketLogger.Logger.error("%@", type: logType, args: "")
```
nuclearace: This should have better logging info.  This should also be log, not error.

Source/SocketIOClient.swift
```diff
@@ -275,6 +275,19 @@ public final class SocketIOClient : NSObject, SocketEngineClient, SocketParsable
...
+        
+        handleEvent("sentPing", data: [""], isInternalMessage: true)
+    }
+    
+    public func engineDidReceivedPong() {
+        DefaultSocketLogger.Logger.error("%@", type: logType, args: "")
```
nuclearace: This should have better logging info. This should also be log, not error.

faizan-carameltech: What other info you require in logs? I will accommodate that. Furthermore closing this PR and opening a new one against development branch with the suggested changes.

PR #666, opened 2017-05-04
Update SocketIOClient.swift

softctrl:
With this change the swift library will have the same behavior of the Java version. When you whant to create a socket client that have namespace on Java you just create a Socket-IO object.

So, if you inform something like "http://your_ip:your_port/path"

It will use your path as a namespace.

### DIFF
--- a/Source/SocketIOClient.swift
+++ b/Source/SocketIOClient.swift
@@ -78,6 +78,10 @@ public final class SocketIOClient : NSObject, SocketEngineClient, SocketParsable
             self.config.insert(.secure(true))
         }
         
+        if !socketURL.path.isEmpty {
+            self.nsp = socketURL.path
+        }
+
         for option in config {
             switch option {
             case let .reconnects(reconnects):


nuclearace:
Is this new behavior in socket.io? I don't remember it doing this before. Namespaces are usually independent of path. Also shouldn't this be looking for a `.path` options supplied in config? That is currently the way you specify custom paths.

What version of socket.io server did you test this in?

nuclearace:
Also this should be opened against development, not master.

softctrl:
sorry for open on the master branch. So about the behavior:

I have a socket io server that have namespace configured.

I have two clients, one for Android and another for iOS.

My android client just do this:

final Socket socket = IO.socket("https://my-ip/my-namespace");

**And it just works well without any other changing on this code.

On the iOS side i just create my socket client in a similar way:

let socket = SocketIOClient(socketURL: URL(string: "https://my-ip/my-namespace",
                                  config: [.log(true), .forcePolling(true)])

So, with that, it not works. I need to change this to:

let socket = SocketIOClient(socketURL: URL(string: "https://my-ip",
                                  config: [.log(true), .forcePolling(true)])
socket.nsp = "/my-namespace"

So, i think that all libraries need to work in the same path. Wy the client Java works right and the Swift code does not?

thanks for your time.

softctrl:
I just miss to send you the versions of the client:

Java -> socket.io-client-0.8.3
Swift -> pod 'Socket.IO-Client-Swift', '~> 8.3.0'

On the server side i am using nodejs with the "socket.io": "1.7.2", but the problem is not here. On the Android project i have an behavior and on the iOS i have another.

nuclearace:
Yeah this was a design choice. I didn't write this client to be like the Java one. The main issue I have with this is that there is another option called `.path` which is different from a namespace, and is actually apart of the url for the socket.io protocol. And I feel like taking the path from the socketURL to be the namespace would bring confusion, which is why I opted to move both options to configurations.

You could also configure the client like:
```swift
let socket = SocketIOClient(socketURL: myURL, config: [.nsp("/mynamespace")])
```

Rather than setting the nsp directly.

softctrl:
ok, so if it is a design choice so thats ok. thanks

softctrl:
sorry my bad. i do not want to close.

meiyongsheng:
How to transfer pictures (larger pictures)

PR #760, opened 2017-07-30
Update void closure usages

abdullahselek:
- With this PR void closures have a better syntax standard.


nuclearace:
I appreciate the PR, but I actually prefer `()` over `Void`. If you look at most Apple code, they use `()`. And actually Void is just defined as [`typealias Void = ()`](https://github.com/apple/swift/blob/master/stdlib/public/core/Policy.swift#L53)

Issue #761, opened 2017-08-02
How can I access SSLSecurity.swift from my Objective-C project?


IbrahimKZ:
I'm working with socket.io in my Objective-C project for 2 years.
Lately I need to integrate SSL pinning. I found this option in 

`[[SocketIOClient alloc] initWithSocketURL:serverURL config:@{... @"secure": [[SSLSecurity alloc] ...}`

but I can't call SSLSecurity class from Objective-C because it doesn't conform to NSObject.
How can I add SSL pinning to socket.io with certificate to my Objective-C project?

ldbfrank:
No, can NOT.

ldbfrank:
It seems the only way is that you should modify file SocketIOClient.swift directly.

IbrahimKZ:
I have found solution using https://github.com/acmacalister/jetfire for Objective-C instead of SSLSecurity.

nuclearace:
Oh, oops. This was one of the reasons I moved Starscream to a fork and out of this project. I think I can fix this.

nuclearace:
Should hopefully be fixed in 11.0.1.

ldbfrank:
Now, you could use Socket.io-client-swift 11.0.1 and Starscream 8.0.2 ! It good!

ldbfrank:
Thanks nuclearace! I have tested, it work!

Issue #598, opened 2016-12-16
SSLCert not visible in Objective-C code


monobono:
When using the SSLCert class: 
`SSLCert *cert = [[SSLCert alloc] initWithData:certData];`

I get the following error:
`Use of undeclared identifier 'SSLCert'`

The Socket.io client is integrated via CocoaPods. All other classes work fine and are visible in Objective-C files.

nuclearace:
I'll fix that. It should be noted however that the WebSocket library I use _is not_ written to be Objective-C compatible. So if you're using things from it, I would highly recommend using Swift.

PR #789, opened 2017-08-29
[WIP] Improve targets

lightsprint09:
This is working in progress branch which has the following goals:

* Remove all targets
* Have one single multi platform target (read [here](http://ilya.puchka.me/xcode-cross-platform-frameworks/) and [here](http://promisekit.org/news/2016/08/Multiplatform-Single-Scheme-Xcode-Projects/))
* Remove Starscream from Source and fetch it via a dependency manager (carthage, spm, pods)

Are you interested in following the this idea?


nuclearace:
Yes, I think this would be great. Granted it doesn't break travis and is still easy to debug.

One note though, I'm using a custom Starscream repo because there were like two changes that had to be made for some ObjC compatibility. Would it be possible to keep that?

nuclearace:
Also for Carthage frameworks. Do you have to include the Checkout? Or is the .resolved and Cartfile enough?

Tests/SocketObjectiveCTest.m
```diff
@@ -99,9 +98,9 @@ - (void)testSocketManager {
     [manager removeSocketWithLabel:@"test"];
 }
 
-- (void)testSSLSecurity {
```
nuclearace: This was one of the custom changes I have to make to Starscream I believe.

nuclearace:
Also, should be opened against `development`

lightsprint09:
Thanks for you feedback.

Maybe I should try to break this up into two separate PRs. One for cross-platform target, the other for eventually removing Starscream code.

Can you give me more information where you did these customization in Starscream

You can set the default branch to development, so all PRs will go against development by default.

lightsprint09:
I close this for now.

nuclearace:
@lightsprint09 I don't actually have access to the settings on this repo, otherwise I would've changed that setting.
https://github.com/nuclearace/Starscream/tree/socket.io is where I keep the changes for this project.

nuclearace:
@lightsprint09 https://github.com/socketio/socket.io-client-swift/issues/761 was one of the reasons I moved Starscream into its own repo. I kept breaking this everytime I synced the local copy of Starscream.

nuclearace:
Also https://github.com/socketio/socket.io-client-swift/issues/598

nuclearace:
I tried to get those changes put into the upstream but the response was Starscream is meant for Swift, so those changes were best left out. 🤷‍♀️

nuclearace:
Also @lightsprint09 it might make more sense to just do it on the Swift 4.0 branch. Since Swift 4 is right around the corner.

lightsprint09:
sure.

PR #796, opened 2017-09-06
Fix Test Coverage iTunes Connect Submission Error

stevethomp:
Xcode 9 now enables Test Coverage on all actions, not only Test. This includes Archive, and iTunes Connect is rejecting binaries with Test Coverage enabled. See more info at https://forums.developer.apple.com/message/242872#242872, https://github.com/Carthage/Carthage/pull/2057#issuecomment-314432344. My uploads are failing with error:
Invalid Bundle - Disallowed LLVM instrumentation. Do not submit apps with LLVM profiling instrumentation or coverage collection enabled. Turn off LLVM profiling or code coverage, rebuild your app and resubmit the app.  

This commit disables "Gather Coverage Data" in the SocketIO-iOS scheme. If coverage is being used we should move to a test w/ coverage scheme, and an Archive scheme with the option disabled.


nuclearace:
This needs to be opened against development.

PR #868, opened 2017-11-16
enabling bitcode

mugx:
would you please enable bitcode?


nuclearace:
@mugx what’s the minimum iOS target bitcode supports?

mugx:
the minimum should be iOS9

PR #889, opened 2017-11-30
Big bug in some linux versions was fixed. For example, in my Ubuntu 1…

vientooscuro:
…6.04.2 and 16.10 with swift 4.0.2 substring of "0" is equal to "4", substring of "[authenticated: true]" is equal to "[authenticated: true"


nuclearace:
I haven't really been working on this branch much anymore. The annoying things you have to get around \@objc stuff kind of ruins it.

PR #915, opened 2017-12-21
Add missing `,`  when leaving or joining a namespace

dgouyette:
if a comma does not follow the namespace, a server will throw : 

play.socketio.protocol.SocketIOEncodingException: Error decoding socket IO packet '0/api/v1/chats/8b52ca9c-aa28-4b5e-b59c-44129d3142eb': Expected ',' to end namespace declaration

As for events, join and leave namespace  should be suffixed by a comma

### DIFF
--- a/Source/SocketIO/Manager/SocketManager.swift
+++ b/Source/SocketIO/Manager/SocketManager.swift
@@ -205,7 +205,7 @@ open class SocketManager : NSObject, SocketManagerSpec, SocketParsable, SocketDa
             return
         }
 
-        engine?.send("0\(socket.nsp)", withData: [])
+        engine?.send("0\(socket.nsp),", withData: [])
     }
 
     /// Called when the manager has disconnected from socket.io.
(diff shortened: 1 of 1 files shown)


nuclearace:
What version of socket.io are you using?

nuclearace:
Because I'm not seeing this version with the latest server version (2.0.4).

dgouyette:
I'm using  https://github.com/playframework/play-socket.io plugin in my server.

Namespace sent to server should have a suffix, like in 

``` private func createPacketString() -> String {
        let typeString = String(type.rawValue)
        // Binary count?
        let binaryCountString = typeString + (type == .binaryEvent || type == .binaryAck ? "\(String(binary.count))-" : "")
        // Namespace?
        let nspString = binaryCountString + (nsp != "/" ? "\(nsp)," : "")
        // Ack number?
        let idString = nspString + (id != -1 ? String(id) : "")

        return completeMessage(idString)
    }``` 
No ?

nuclearace:
This is more of an issue of their server implementation. Join packets don't need to have a trailing ",". It works fine with the reference JS implementation.

dgouyette:
Thanks.

PR #927, opened 2018-01-10
fix insert invoked twice

slidoooor:
unction self._config.insert was invoked twice in the init() function,once in the init() function,and the other was in the setConfig() function ^-^


nuclearace:
PRs should be opened against the development branch.

nuclearace:
Manually merged into development.

PR #993, opened 2018-03-30
v13.2.0

nuclearace:
- Add ability to bypass Data inspection in emits. #992


opfeffer:
This looks cool -- is there an ETA when 3.2.0 might get released?
Thanks for your work!

nuclearace:
@opfeffer Should be sometime this week I hope.

PR #1014, opened 2018-05-09
thread-safety on ackNumber generation and memory leak fix

alevinru:
1. ack number generation without semaphores is failing under heavy load (1000+ emits from a concurrent queue)
2. without weak self in timeout callback it retains the callbacks if timeout > 0


nuclearace:
Also, you a mention concurrent queue. The library is not designed to handle concurrent queues. https://github.com/socketio/socket.io-client-swift/blob/development/Source/SocketIO/Manager/SocketManager.swift#L89. I guess I forgot to update the documentation for the handleQueue option.

alevinru:
Now it's safe to use concurrent queue

Before the fix i've tried to use main queue and there were a lot of crashes nevertheless

PR #1015, opened 2018-05-09
thread-safe ackNumber generation and memory leak fix

alevinru:
1. ack number generation without semaphores is failing under heavy load (1000+ emits from a concurrent queue)
2. without weak self in timeout callback it retains the callbacks if timeout > 0

### DIFF
--- a/Source/SocketIO/Ack/SocketAckEmitter.swift
+++ b/Source/SocketIO/Ack/SocketAckEmitter.swift
@@ -135,10 +135,14 @@ public final class OnAckCallback : NSObject {
 
         guard seconds != 0 else { return }
 
-        socket.manager?.handleQueue.asyncAfter(deadline: DispatchTime.now() + seconds) {[weak socket] in
+        socket.manager?.handleQueue.asyncAfter(deadline: DispatchTime.now() + seconds) {
+            [weak self, weak socket] in
+            
             guard let socket = socket else { return }
-
-            socket.ackHandlers.timeoutAck(self.ackNumber)
+            guard let me:OnAckCallback = self else { return }
+            
+            socket.ackHandlers.timeoutAck(me.ackNumber)
+            
         }
     }
 
--- a/Source/SocketIO/Ack/SocketAckManager.swift
+++ b/Source/SocketIO/Ack/SocketAckManager.swift
@@ -60,18 +60,27 @@ private struct SocketAck : Hashable {
 
 class SocketAckManager {
     private var acks = Set<SocketAck>(minimumCapacity: 1)
+    private let ackSemaphore = DispatchSemaphore(value: 1)
 
     func addAck(_ ack: Int, callback: @escaping AckCallback) {
+        ackSemaphore.wait()
+        defer { ackSemaphore.signal() }
         acks.insert(SocketAck(ack: ack, callback: callback))
     }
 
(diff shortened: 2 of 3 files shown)


nuclearace:
@alevinru As mentioned in the previous PR, this library is designed around the principal that it should only work around serial queues, and indeed it _can_ only really work with serial queues because of the way that socket.io works. This would be one change that would be needed to make it work with serial queues, but there would need to be more.

For example, adding handlers vs executing them would need to locked, managing the binary data that socket.io sends back (which is the part that would start breaking with concurrent queues) would need to be locked. And there's probably others that would also need to be change.

When I wrote this library it originally had a pretty complex queue system, and I had hoped to make it work with concurrent queues, but it quickly became apparent that the complexity, and how socket.io behaves in JS and the other clients, made this not feasible.

alevinru:
@nuclearace Ok, but i'm not about the internal queue of the library. I have an event bus inside my app and it concurrently emits with ack to socket.io. Then the library generates ack numbers, and only after it pushes operations into the internal library's serial queue. That is why the ack generation have to be thread-safe.

I've tried the library as it is without the semaphores and it crashes a lot because some simultaneous emits receive the same ack numbers.

nuclearace:
@alevinru The `handleQueue` is not really the internal queue in this case. It's the queue that all interaction with the manager/client should be done, and the queue must be a serial queue.

alevinru:
@nuclearace Do you mean i shouldn't just emit from any part of my code, but have to dispatch a block operation to the SocketManager's handleQueue?

nuclearace:
@alevinru Yes exactly. There's a note on that in the docs for the client. https://nuclearace.github.io/Socket.IO-Client-Swift/Classes/SocketIOClient.html. I'll add a note for it for the manager as well.

I'm going to close this PR. I appreciate the work, but this isn't something I think the library can or should support.

alevinru:
@nuclearace This is quite strange not to merge this PR, because it's brings nothing negative, just a more convenient way to use the library. And the callback retained by self reference is still leaking regardless of the serial or concurrent question

nuclearace:
@alevinru It's still incorrect to use a concurrent queue as the handleQueue regardless. Because of the way socket.io sends binary, packets must be processed in sequence. A concurrent queue would make it impossible to correctly attach binary to packets without adding more queues. Which means more locks and dispatching. 

The strong capture can be its own PR.

alevinru:
@nuclearace I don't use a concurrent queue as the handleQueue - It's serial and i understand why it is required. 

But i don't understand why i need to wrap every call to emit:withAck: into a dispatch-to handleQueue block?

I.e. instead of just:
```
[self.socket emitWithAck:@"event" with:@[@payload]];
```

We need to remember somewhere the socket handleQueue instance and:
```
    dispatch_async(socketHandleQueue, ^{
        [self.socket emitWithAck:@"event" with:@[@payload]];
    });
```

Even if someone has configured SocketManager instance with default main queue as handleQueue, the app may be in background at the time of the emit - then the call to emitWithAck:with: will be dispatched from some background queue, not the SocketManager's handleQueue.

As i see in your code there's no need to demand this dispatching to socketHandleQueue before emitting because all the packets are then processed serially from the serial handleQueue.

The only issue is ack number generation and acks Set mutation after receiving ack.

PR #1034, opened 2018-06-07
Fixed loosing POST packets on web socket connection failure

OneSman7:
Fixed loosing POST packets on web socket connection failure and engine memory leak on server errors.

### DIFF
--- a/Source/SocketIO/Engine/SocketEngine.swift
+++ b/Source/SocketIO/Engine/SocketEngine.swift
@@ -371,6 +371,9 @@ open class SocketEngine : NSObject, URLSessionDelegate, SocketEnginePollable, So
         fastUpgrade = false
         probing = false
         flushProbeWait()
+        if postWait.count != 0 {
+            flushWaitingForPostToWebSocket()
+        }
     }
 
     private func flushProbeWait() {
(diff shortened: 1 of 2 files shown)


Source/SocketIO/Engine/SocketEngine.swift
```diff
@@ -371,6 +371,9 @@ open class SocketEngine : NSObject, URLSessionDelegate, SocketEnginePollable, So
         fastUpgrade = false
         probing = false
         flushProbeWait()
+        if postWait.count != 0 {
```
nuclearace: How about a `!.isEmpty`?

OneSman7: No problem, I just copied existing code.

nuclearace: Or maybe move that check into flush... and unconditionally call it?

nuclearace: I'm fine with this way in that case! Would you mind also adding some documentation comments to these methods that explain this failure case? I think those will be helpful when I come back in 6 months and try to refactor and end up breaking it again :).

OneSman7: 👌 will also change to `!.isEmpty` and test

PR #1080, opened 2018-09-03
fix cookies import when create the websocket

hschouman:
The upgrade to web socket should import only cookies relative to the url.
Without this specify, all cookies are imported and in my case it never upgrade to web socket (because there are several cookies in header that does not match the session)

### DIFF
--- a/Source/SocketIO/Engine/SocketEngine.swift
+++ b/Source/SocketIO/Engine/SocketEngine.swift
@@ -281,7 +281,7 @@ open class SocketEngine : NSObject, URLSessionDelegate, SocketEnginePollable, So
     private func createWebSocketAndConnect() {
         var req = URLRequest(url: urlWebSocketWithSid)
 
-        addHeaders(to: &req, includingCookies: session?.configuration.httpCookieStorage?.cookies)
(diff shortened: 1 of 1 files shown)


hschouman:
For more information on my use case with cookies here, I use session affinity on heroku :
https://devcenter.heroku.com/articles/session-affinity

But I guess there are other use cases with cookies here

nuclearace:
Hm, the idea for this was to allow cookies from the initial polling connection to carry over to the websocket connection.

leuducquy:
Merge

nuclearace:
@hschouman Have you tested if this ever actually returns cookies? There's a scheme difference between `urlWebSocketWithSid` and `urlPollingWithSid`, and session will only ever have the polling url.

hschouman:
You are right. In my case I have the same result with the two schemes because the cookies are stored with the base url (path == "/").
I've just fixed that.

nuclearace:
@hschouman One last thing, can you change this to be opened against development?

OneSman7:
Found this after made comment in my PR #1058 (https://github.com/socketio/socket.io-client-swift/pull/1058#issuecomment-425062388).
This PR also fixes bug described there. It is only left to decide whether it will be better to pass filtered cookies to _addHeaders_ method or to filter them by domain inside _addHeaders_ method.

PR #1097, opened 2018-09-29
Add optional write completion handler for emit's

### LINKED ISSUE #1096: Feature Request: completion handler for writes
We've hit a case where we need to know when a write to the underlying socket has been completed. We're sending quite a large amount of latency sensitive data, such that the network can sometimes not keep up. It would be really helpful to be able to monitor when a write actually completes so we can scale back writes when the network slows down. The underlying Starscream WebSokcet class already provides this functionality via a completion handler called on write completion. Ideally this would be exposed up through socket.io as well.

We've considered using the ack mechanism from the server that's already supported by this library, but the extra latency that the network call introduces is unacceptable for our use case. 

Is there a way to achieve this with the current APIs? And if not would you accept a PR adding this functionality?

headlessme:
A first draft to add an optional completion handler for calls to emit, see #1096 for reasoning. I'm not a Swift expert so I'm quite likely doing something dumb here :)

Thanks to @AgentFeeble for providing a great starting point. 

Fixes #1096

### DIFF
--- a/Source/SocketIO/Client/SocketIOClient.swift
+++ b/Source/SocketIO/Client/SocketIOClient.swift
@@ -231,7 +231,14 @@ open class SocketIOClient : NSObject, SocketIOClientSpec {
     /// - parameter items: The items to send with this event. May be left out.
     /// - parameter completion: Callback called on transport write completion.
     open func emit(_ event: String, _ items: SocketData..., completion: @escaping () -> ())  {
-        emit([event] + items, completion: completion)
+        do {
+            try emit(event, with: items.map({ try $0.socketRepresentation() }), completion: completion)
+        } catch {
+            DefaultSocketLogger.Logger.error("Error creating socketRepresentation for emit: \(event), \(items)",
+                                             type: logType)
+
+            handleClientEvent(.error, data: [event, items, error])
+        }
     }
 
     /// Same as emit, but meant for Objective-C
--- a/Source/SocketIO/Client/SocketIOClient.swift
+++ b/Source/SocketIO/Client/SocketIOClient.swift
@@ -213,7 +213,7 @@ open class SocketIOClient : NSObject, SocketIOClientSpec {
     /// - parameter items: The items to send with this event. May be left out.
     open func emit(_ event: String, _ items: SocketData...) {
         do {
-            try emit(event, with: items.map({ try $0.socketRepresentation() }))
+            try emit(event, with: items.map({ try $0.socketRepresentation() }), completion: {})
         } catch {
             DefaultSocketLogger.Logger.error("Error creating socketRepresentation for emit: \(event), \(items)",
                                              type: logType)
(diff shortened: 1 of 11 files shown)


nuclearace:
This looks pretty good! Can you open this against the development branch?

headlessme:
Great! I've changed the base branch

nuclearace:
Excellent, one thought I have though: Would it make sense to wrap the user provided callback with one that wraps and dispatches on the `handleQueue`? If we don't do that we'd need to document on the method that the provided closure won't be executed on the handle queue.

nuclearace:
Something maybe like

```swift
{[weak self] in 
  guard let this = self else { return }
  this.handleQueue.async { /* call callback */ }
}

**Comment by headlessme** [comment]
Good idea – added!

File: Source/SocketIO/Client/SocketIOClient.swift [hunk 1]
```diff
@@ -295,8 +295,14 @@ open class SocketIOClient : NSObject, SocketIOClientSpec {
     }
 
     func emit(_ data: [Any], ack: Int? = nil, binary: Bool = true, isAck: Bool = false, completion: (() -> ())? = nil) {
+        // wrap the completion handler so it always runs async via handlerQueue
+        let wrappedCompletion = {[weak self, completion] in
```
nuclearace: You shouldn't need to put completion in the capture list.

headlessme: Fixed

Source/SocketIO/Client/SocketIOClient.swift
```diff
@@ -228,7 +228,17 @@ open class SocketIOClient : NSObject, SocketIOClientSpec {
...
+        emit([event] + items, completion: {})
+    }
+    
+    /// Same as emit, but meant for Objective-C
+    ///
+    /// - parameter event: The event to send.
+    /// - parameter items: The items to send with this event. Send an empty array to send no data.
+    /// - parameter completion: Callback called on transport write completion.
+    @objc
+    open func emit(_ event: String, with items: [Any], completion: @escaping () -> ()) {
```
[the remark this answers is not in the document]
headlessme: Added a method for Swift. I also added the method signature to the SocketIOClientSpec, but not sure if that's needed?

Source/SocketIO/Client/SocketIOClient.swift
```diff
@@ -213,22 +213,44 @@ open class SocketIOClient : NSObject, SocketIOClientSpec {
...
+    /// Send an event to the server, with optional data items and write completion handler.
+    ///
+    /// If an error occurs trying to transform `items` into their socket representation, a `SocketClientEvent.error`
+    /// will be emitted. The structure of the error data is `[eventName, items, theError]`
+    ///
+    /// - parameter event: The event to send.
+    /// - parameter items: The items to send with this event. May be left out.
+    /// - parameter completion: Callback called on transport write completion.
+    open func emit(_ event: String, _ items: SocketData..., completion: @escaping () -> ())  {
+        emit([event] + items, completion: completion)
```
nuclearace: This needs to do the transformation from SocketData to something representable in JSON. See the other swift-centric methods.

headlessme: Ah yea of course, added.

headlessme:
Great! Thanks for the quick review turnaround 👍

PR #1102, opened 2018-10-08
Propagate http header callback from websocket back to client.

vonox7:
On Socket.IO Android we already have this callback, so backport it to iOS.

### DIFF
--- a/Source/SocketIO/Client/SocketIOClientSpec.swift
+++ b/Source/SocketIO/Client/SocketIOClientSpec.swift
@@ -334,4 +334,15 @@ public enum SocketClientEvent : String {
     /// }
     /// ```
     case statusChange
+
+    /// Emitted when when upgrading the http connection to a websocket connection.
+    ///
+    /// Usage:
+    ///
+    /// ```swift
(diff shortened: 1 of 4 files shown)


vonox7:
Note: we need the current (unreleased) Starscream version, as we need the newly introduced callback from there.

nuclearace:
@vonox7 Can you open this against development?

Source/SocketIO/Client/SocketIOClientSpec.swift
```diff
@@ -334,4 +334,15 @@ public enum SocketClientEvent : String {
...
+    /// Emitted when when upgrading the http connection to a websocket connection.
+    ///
+    /// Usage:
+    ///
+    /// ```swift
+    /// socket.on(clientEvent: .httpResponseHeaders) {data, ack in
+    ///     // Some header logic
+    /// }
+    /// ```
+    case httpResponseHeaders
```
nuclearace: If this is notifying when switching transports, it should be named something `websocketUpgrade` and the payload documented.

vonox7: you're right, fixed in next PR

PR #1103, opened 2018-10-08
Propagate http header callback from websocket back to client.

vonox7:
On Socket.IO Android we already have this callback, so backport it to iOS.
Also note that we need the current (unreleased) version of Starscream (master branch commit 0f18ae5959b2ff9940b5bc50a4f770f4c38bbd2f) to run it.


vonox7:
@nuclearace Starscream released now this feature :)

PR #1110, opened 2018-10-17
v13.4.0

nuclearace:
- Add emits with write completion handlers.
- Add ability to listen for when a websocket upgrade happens


headlessme:
@nuclearace any idea when this release will happen? Don't want to rush you, but we're waiting on it to get a new SDK out of the door 😄 Thanks!

Issue #1118, opened 2018-11-07
Optional write completion handler for emit's cause major crashes


vonox7:
PR https://github.com/socketio/socket.io-client-swift/pull/1097 creates major crashes in production. When reverting the PR (see https://github.com/vonox7/socket.io-client-swift/commits/v-revert-completion-handler ) the crashes are gone.

Crashed: NSOperationQueue 0x280ab42e0 (QOS: UNSPECIFIED)
EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000a35918130

Stacktrace:
```
Crashed: NSOperationQueue 0x280ab42e0 (QOS: UNSPECIFIED)
0  libobjc.A.dylib                0x1d853cd70 objc_msgSend + 16
1  Starscream                     0x1050ff6a8 protocol witness for WSStream.write(data:) in conformance FoundationStream (WebSocket.swift:241)
2  Starscream                     0x105105c30 closure #1 in WebSocket.dequeueWrite(_:code:writeCompletion:) (WebSocket.swift:1268)
3  Starscream                     0x1050fd244 thunk for @escaping @callee_guaranteed () -> () (<compiler-generated>)
4  Foundation                     0x1d9de3b6c __NSBLOCKOPERATION_IS_CALLING_OUT_TO_A_BLOCK__ + 16
5  Foundation                     0x1d9cebcc8 -[NSBlockOperation main] + 72
6  Foundation                     0x1d9ceb19c -[__NSOperationInternal _start:] + 740
7  Foundation                     0x1d9de5a40 __NSOQSchedule_f + 272
8  libdispatch.dylib              0x1d8d8f6c8 _dispatch_call_block_and_release + 24
9  libdispatch.dylib              0x1d8d90484 _dispatch_client_callout + 16
10 libdispatch.dylib              0x1d8d66e14 _dispatch_continuation_pop$VARIANT$armv81 + 404
11 libdispatch.dylib              0x1d8d664f8 _dispatch_async_redirect_invoke + 592
12 libdispatch.dylib              0x1d8d72afc _dispatch_root_queue_drain + 344
13 libdispatch.dylib              0x1d8d7335c _dispatch_worker_thread2 + 116
14 libsystem_pthread.dylib        0x1d8f72190 _pthread_wqthread + 472
15 libsystem_pthread.dylib        0x1d8f74d00 start_wqthread + 4
```

[...]

PR #1119, opened 2018-11-07
Don’t create empty closures when we don’t need them

vonox7:
Improves performance, fixes https://github.com/socketio/socket.io-client-swift/issues/1118.

### DIFF
--- a/Source/SocketIO/Client/SocketIOClient.swift
+++ b/Source/SocketIO/Client/SocketIOClient.swift
@@ -230,7 +230,7 @@ open class SocketIOClient : NSObject, SocketIOClientSpec {
     /// - parameter event: The event to send.
     /// - parameter items: The items to send with this event. May be left out.
     /// - parameter completion: Callback called on transport write completion.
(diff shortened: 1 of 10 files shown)


Source/SocketIO/Client/SocketIOClient.swift
```diff
@@ -230,7 +230,7 @@ open class SocketIOClient : NSObject, SocketIOClientSpec {
     /// - parameter event: The event to send.
     /// - parameter items: The items to send with this event. May be left out.
     /// - parameter completion: Callback called on transport write completion.
-    open func emit(_ event: String, _ items: SocketData..., completion: @escaping () -> ())  {
+    open func emit(_ event: String, _ items: SocketData..., completion: (() -> ())? = nil)  {
```
nuclearace: I think this one can stay nil right? If you're calling this method then you're explicitly wanting to use the callback feature.

vonox7: You are right. I removed the unnecessary function calls.

PR #1123, opened 2018-11-19
Convert to Swift 4.2

DmitryPR:
Hello,

I have converted the current socket 1.0 branch to Swift version 4.2 

Could this get merged and released as 9.0.2 tag to the cocoapods?

I am using your library as s.dependecy in the podfile and unfortunately cannot specify branch in there.
https://github.com/salemove/ios-bundle/blob/master/SalemoveSDK.podspec#L26


Source/SocketStringReader.swift
```diff
@@ -40,33 +40,34 @@ struct SocketStringReader {
...
-        let readString = message[currentIndex..<message.index(currentIndex, offsetBy: count)]
+        let readString = message[currentIndex..<message.characters.index(currentIndex, offsetBy: count)]
+        
         advance(by: count)
         
         return String(readString)
     }
     
     mutating func readUntilOccurence(of string: String) -> String {
-        let substring = message.utf16[currentIndex...]
```
nuclearace: Have you verified this with unicode characters and socket.io 1.0?

DmitryPR: Think I got these changes from your update to Swift 4.0 but I reverted them and just updated the Swift version and NSData -> Data change.

nuclearace:
Have you verified this complies? I'm getting an error for not unwrapping an optional.

DmitryPR:
I have verified it compiles, what optional is not unwrapped?

DmitryPR:
Also I noticed the initial swift4 branch has 11 tests failing, is this intended?

Source/SocketExtensions.swift
```diff
@@ -31,7 +31,11 @@ enum JSONError : Error {
 
 extension Array {
     func toJSON() throws -> Data {
-        return try JSONSerialization.data(withJSONObject: self, options: JSONSerialization.WritingOptions(rawValue: 0))
+        if #available(iOSApplicationExtension 11.0, *) {
```
nuclearace: I think you might need more checks here. Does this compile for tvOS/macOS targets?

DmitryPR: You are correct, added the fix

Source/SocketExtensions.swift
```diff
@@ -31,7 +31,21 @@ enum JSONError : Error {
 
 extension Array {
     func toJSON() throws -> Data {
-        return try JSONSerialization.data(withJSONObject: self, options: JSONSerialization.WritingOptions(rawValue: 0))
+        #if os(iOS)
+        if #available(iOS 11.0, *) {
```
nuclearace: You should be able to specify multiple things here. `#available(iOS 11.0, macOS 10.13, *)`

PR #1126, opened 2018-11-28
Swift update

DmitryPR:
Bump Swift version in podspec to 4.2 

Update Starscream to 3.0.6


nuclearace:
Why is this needed? My previous experience with Xcode is that you can leave the Swift version at the last major version and it will pick up the newest compatible version.

nuclearace:
Also this would need to be opened on the development branch.

DmitryPR:
As a framework developer I encountered an issue that the dependencies that I am using are not having the swift-version defined. 

This leads to 2 problems:

1) The podspec should really specify what swift version it has been adapted to, so the integrators know in advance what to expect when trying to solve different version problem

2) https://github.com/CocoaPods/cocoapods-packager/issues/210

nuclearace:
Okay, just to play it as safe as possible, I'm going to go ahead and release 13.4.0 now before merging this. And this can be the starting point for v14.0.0. (Really the last release should have probably been 14 since Starscream actually broke semver in the last release).

PR #1149, opened 2019-01-12
Add configurable exponential backoff

headlessme:
Added a basic exponential backoff algorithm to the reconnect logic. It support jittering via `randomizationFactor` factor, and limiting the max interval via `reconnectWaitMax`.


nuclearace:
Nice! Could you do me a favor, and add an entry for `v14.0.0` in the changelog detailing this addition? It just so happened I was about to release a new one soon.

headlessme:
How does that look?

PR #1179, opened 2019-03-28
v15.1.0

### LINKED ISSUE #1178: Socket emit completion handler not called
55056149-ef80f300-5097-11e9-8fea-7b323d4ca329

It didn't step to the completion handler. When I debugged it (step by step slowly), it get called. I don't know the reason, please check it.

nuclearace:
- Fix #1178


nuclearace:
@peihsendoyle Yes that commit should fix the issue. If you want to test that out, that'd be great.

peihsendoyle:
> @peihsendoyle Yes that commit should fix the issue. If you want to test that out, that'd be great.

yes, I've checked it. It really works. Thanks very much.

PR #1198, opened 2019-05-17
Expose SocketIOClient and SocketAnyEvent to objective c

minlite:
This helps expose SocketIOClient and SocketAnyEvent members to Obj-C.


nuclearace:
What issue were you facing? AFAIU, this shouldn't be needed.

minlite:
I could not see/use any of the methods and properties from the library in Objective C.

nuclearace:
How were you importing the library?

PR #1214, opened 2019-07-05
Auto detect which protocol to use depend on url scheme

JadianZheng:


Source/SocketIO/Engine/SocketEngine.swift
```diff
@@ -250,6 +250,20 @@ open class SocketEngine : NSObject, URLSessionDelegate, SocketEnginePollable, So
         if client == nil {
             return (URL(string: "http://localhost/")!, URL(string: "http://localhost/")!)
         }
+        
+        if let scheme = URLComponents(string: url.absoluteString)!.scheme {
```
nuclearace: I feel like there should be a way of doing this using the `url`, or by using `?` instead of `!`.

JadianZheng: Thank you for your patient. I will fix it soon.

Issue #1016, opened 2018-05-10
Reconnection doesn't work


kvaDrug:
I use Socket IO v13.1.1 on iOS app with server (apache nginx, Socket IO v2.0.4). We test what will happen if server reboots. 

1. iOS app is connected to the server
2. Then the server is rebooted
3. iOS app tries to reconnect

iOS library constantly emits `reconnectAttempt` event (default config). Why it doesn't reconnects?
Is iOS app supposed to perform some custom actions?

kvaDrug:
I was able to fix my problem by disabling built-in reconnection:
    manager = SocketManager(socketURL: serverURL, config: [.reconnects(false)])
And then I created a `Timer` that will periodically check if socket is connected and reconnects if necessary.

wongzigii:
I have the same problem

nuclearace:
Sorry for missing this. Does the logs spit out something about an unknown session id?

nickngsr:
I have the same issue. No logs seem to spit out about it at all.

If I listen to the 'reconnectAttempt' event, which are still fired as expected, and in that I call disconnect on the SocketManager, the reconnect attempt will work.

nickngsr:
It would appear if a reconnect attempt fails it doesn't close the socket for the next reconnect attempt

```
2018-06-18 18:29:05.424781+0100 four[2661:2628085] LOG SocketManager: Starting reconnect
2018-06-18 18:29:05.425002+0100 four[2661:2628085] LOG SocketIOClient{/ipad}: Handling event: statusChange with data: [connecting]
2018-06-18 18:29:05.425093+0100 four[2661:2628085] LOG SocketIOClient{/ipad}: Handling event: reconnect with data: ["Socket Disconnected"]
Reconnect
2018-06-18 18:29:05.425401+0100 four[2661:2628085] LOG SocketManager: Trying to reconnect
2018-06-18 18:29:05.425491+0100 four[2661:2628085] LOG SocketIOClient{/ipad}: Handling event: reconnectAttempt with data: [-1]
Reconnect Attempt
2018-06-18 18:29:05.437001+0100 four[2661:2628085] LOG SocketManager: Adding engine
2018-06-18 18:29:05.437254+0100 four[2661:2628112] LOG SocketEngine: Engine is being released
2018-06-18 18:29:05.437425+0100 four[2661:2628251] LOG SocketEngine: Starting engine. Server: https://xxxx
[...]

Issue #918, opened 2018-01-02
Socket disconnects automatically, reconnects, and disconnects again and form a loop.


kinza88:
Hey i am using socket.io version 13.1.0 (latest)
When app in launched, after some (random) time socket gets disconnected. It reconnects again and then again it gets disconnected and thus form a loop.
My code is given as follows:

```
let manager = SocketManager(socketURL: URL(string: "https://staging.discretelogix.com:5104/")!, config: [.compress, .log(true), .reconnects(true) ])
let socket = manager.defaultSocket
socket.connect()
socket.on("connect") {data, ack in
      print ("Socket connected")
}
```

Every time socket gets disconnected, one of the following reasons appear in logs:
1) Could not connect to server
2) Connection closed by server
3) Session id unknown. 

I have also tried config options: forceNew(true), forcePolling(true), reconnects(true) but the problem didn't get resolved. 

@nuclearace your help would be appreciated. 
Thankyou.

bl82:
Same for me, my SocketIO init:
`        self.manager = SocketManager(socketURL: URL(string: str)!, config: [.log(true), .connectParams(["user_id": id, "room_code": roomId])])
        self.socket = manager.defaultSocket
        self.socket.on("connect") {data, ack in
            
            let dict = [["sid":self.socket.sid]]
            
            loggerHandler(dict, "connected")
            print("socket connected")
        }`

ConnectDevVS:
Im facing this issue on 'Socket.IO-Client-Swift', '~> 13.1.0' . did anyone have a solution for this?
@bl82 : were you able to find a solutions for this

bl82:
@ConnectDevVS Yes, we made a code review in our team, traced whole flow with backend. Bug came from out side. During incorrect flow connect/reconnect/disconnect sequence.

ConnectDevVS:
@bl82 : Thanks for the rely
Also, can you give me some more idea of what went wrong or what you guys did that gave rise to this issue. Im trying to figure what measure to resolve this and this info will be of great help. We are using a simple node server, and our android counter part were able to connect to it seamlessly.

When "forceNew" parameter is given in the configuration, the iOS app was able to reconnect back after disconnection, without this parameter, when connection is lost, it just continue to reconnect without success.

bl82:
[...]

Issue #913, opened 2017-12-19
Socket disconnects automatically and after disconnect it's not reconnect. Get lots of logs in console.


bhautikmobileappdev:
I tried to reconnect but got following logs

Got event: statusChange, with items: Optional([SocketIO.SocketIOClientStatus])
Got event: reconnect, with items: Optional(["Ping timeout"])
Got event: reconnectAttempt, with items: Optional([-1])
Got event: statusChange, with items: Optional([SocketIO.SocketIOClientStatus])
Got event: statusChange, with items: Optional([SocketIO.SocketIOClientStatus])
Got event: reconnectAttempt, with items: Optional([-2])
Got event: statusChange, with items: Optional([SocketIO.SocketIOClientStatus])
Got event: statusChange, with items: Optional([SocketIO.SocketIOClientStatus])
Got event: reconnectAttempt, with items: Optional([-3])
Got event: statusChange, with items: Optional([SocketIO.SocketIOClientStatus])
Got event: disconnect, with items: Optional(["Got Disconnect"])
Got event: error, with items: Optional(["Session ID unknown"])

Please check and let me know if i did anything wrong.

Thanks in advance.

nuclearace:
Can you provide more detail? What version are you using on the client/server, what configs are used, server settings (ping info.)

bhautikmobileappdev:
pod 'Socket.IO-Client-Swift', '10.0.1'

kinza88:
Hey.. did you get any solution to this problem?

bl82:
Same problem for me I constantly get "Session ID unknown" error:
My Podfile:
```
source 'https://github.com/CocoaPods/Specs.git'
platform :ios, '8.4'
use_frameworks!
target 'Votecast' do
pod 'Socket.IO-Client-Swift' # Or latest version
pod 'Fabric'
pod 'Crashlytics'
pod 'MBProgressHUD'
pod 'PNChart'

end
```
My SocketIO swift class:
```
import Foundation
import SocketIO
@objc public class SocketProvider : NSObject {
    
    let str = "https://dev.votecast.dk/" //Prod server
    
    var descr = "SocketProvider class"
    var socket : SocketIOClient
    var manager : SocketManager
```
Init of Socket Manager:
```
        self.manager = SocketManager(socketURL: URL(string: str)!, config: [.log(true), .connectParams(["user_id": id])])
        self.socket = manager.defaultSocket
        self.socket.on("connect") {data, ack in
            
[...]

Issue #909, opened 2017-12-17
After connection upgrade to webSocket In case of a failure to reconnect client will never reconnect again and be stuck in an endless reconnect loop


odedsh:
I am using a server behind a NGINX load balancer. My server had an unexpected shutdown and restart. It takes a little time for the server to complete the boot process and during that time NGINX would return an error to anyone who tries to connect to the server.

It seems that if the server fails to reconnect once then it will not try again:

Setup to reproduce:

SocketManager connects initially then upgrades the connection to a websocket 
The server on the other side shuts down causing the socket to close and the client code 
 It takes the server ~30 seconds to come back and be available

It seems that in the code.
After the Socket connection was established the state is changed in ```SocketEngine::doFastUpgrade``` to polling = false.

Now when the code finds out that status is disconnected it will start to Reconnect by calling connect
```
    /// Override if you wish to attach a custom `SocketEngineSpec`.
    open func connect() {
        guard !status.active else {
            DefaultSocketLogger.Logger.log("Tried connecting an already active socket", type: SocketManager.logType)

            return
        }

        if engine == nil || forceNew {
            addEngine()
        }

        status = .connecting

        engine?.connect()
    }
```

The engine call to connect will call doLongPoll however since polling is false only a single request will be performed.
This action will generate a single error which will not update the status flag and now all future calls to reconnect will fail at the guard above. And since we are not polling no additional calls will be made to the server and the system will stay forever in status .connecting

Logs:

Dec 17 10:26:21 Gils-MacBook-Pro-4 wisdo[96407]: LOG SocketManager: Starting reconnect
Dec 17 10:26:21 Gils-MacBook-Pro-4 wisdo[96407]: LOG SocketIOClient{/}: Handling event: statusChange with data: [connecting]
Dec 17 10:26:21 Gils-MacBook-Pro-4 wisdo[96407]: LOG SocketIOClient{/}: Handling event: reconnect with data: ["Socket Disconnected"]
Dec 17 10:26:21 Gils-MacBook-Pro-4 wisdo[96407]: LOG SocketManager: Trying to reconnect
Dec 17 10:26:21 Gils-MacBook-Pro-4 wisdo[96407]: LOG SocketIOClient{/}: Handling event: reconnectAttempt with data: [-1]
Dec 17 10:26:21 Gils-MacBook-Pro-4 wisdo[96407]: LOG SocketEngine: Starting engine. Server: wss://dev1.wisdo.com
[...]

PR #1231, opened 2019-09-07
Fixed endless reconnection loop

salpeev:
Fix for:
https://github.com/socketio/socket.io-client-swift/issues/1016
https://github.com/socketio/socket.io-client-swift/issues/918
https://github.com/socketio/socket.io-client-swift/issues/913
https://github.com/socketio/socket.io-client-swift/issues/909
https://github.com/socketio/socket.io-client-swift/issues/784


nuclearace:
How exactly is this fixing it? I guess not dispatching again to the engineQueue is better, but I'm not seeing how this would fix anything per se.

salpeev:
I can't tell you exact sequence of events right now, but the basic idea is as follows. Method '_connect' is executed on serial 'engineQueue'.  'disconnect' is called at the beginning of '_connect' method and dispatches '_disconnect' to the same 'engineQueue'. So '_disconnect' isn't executed immediately, but is being dispatched for a later execution. After that '_connect' method continues its execution (probably creates some connection-related stuff). When '_connect' returns, previously dispatched '_disconnect' is being executed (and probably kills some connection-related stuff created by '_connect'). So calling '_disconnect' instead of 'disconnect' ensures proper ordering of events

PR #1233, opened 2019-09-22
Fix connection error handling if the connection state is wrong

allenlinli:
- fix: handle failure in connect
- feat: add test `testConnectedAlready`


nuclearace:
@allenlinli Looks like that test failed

allenlinli:
I see, I'll fix it in the next week.

allenlinli:
Sorry for the late update.

Perronef5:
Hi can we merge this into develop?

PR #1324, opened 2021-02-05
Added client emit methods with array parameters

OneSman7:
This PR mitigates  an issue with your code emit methods syntax, which breaks the use of it if wrapped in another method.
Take for example a simple emit:

open func emit(_ event: String, _ items: SocketData..., completion: (() -> ())? = nil)  {
        do {
            emit([event] + (try items.map({ try $0.socketRepresentation() })), completion: completion)
        } catch {
            DefaultSocketLogger.Logger.error("Error creating socketRepresentation for emit: \(event), \(items)",
                                             type: logType)

            handleClientEvent(.error, data: [event, items, error])
        }
    }
You make use of variadic parameter, which is totally fine when used directly. For example socket.emit(event, arg1, arg2)

But if you wrap the call to this method:

func send(event: String, with arguments: [SocketTransferable]) {
       socket.emit(event, arguments)
}
the call send(event, arguments) will result in arguments being wrapped in array again because of how variadic parameters work. I can even write my method like this

func send(event: String, with arguments: SocketTransferable...) {
       socket.emit(event, arguments)
}
and still get errors from server for calls like send(event, arg1, arg2).

It adds methods overloads that accept parameters as array.

### DIFF
--- a/Source/SocketIO/Client/SocketIOClient.swift
+++ b/Source/SocketIO/Client/SocketIOClient.swift
@@ -212,6 +212,19 @@ open class SocketIOClient: NSObject, SocketIOClientSpec {
     /// - parameter items: The items to send with this event. May be left out.
     /// - parameter completion: Callback called on transport write completion.
     open func emit(_ event: String, _ items: SocketData..., completion: (() -> ())? = nil)  {
+        emit(event, items, completion: completion)
+    }
+    
+    /// Send an event to the server, with optional data items and optional write completion handler.
+    ///
+    /// If an error occurs trying to transform `items` into their socket representation, a `SocketClientEvent.error`
+    /// will be emitted. The structure of the error data is `[eventName, items, theError]`
+    ///
+    /// - parameter event: The event to send.
+    /// - parameter items: The items to send with this event. May be left out.
+    /// - parameter completion: Callback called on transport write completion.
+    open func emit(_ event: String, _ items: [SocketData], completion: (() -> ())?) {
+        
         do {
             emit([event] + (try items.map({ try $0.socketRepresentation() })), completion: completion)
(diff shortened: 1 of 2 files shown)


Source/SocketIO/Client/SocketIOClient.swift
```diff
@@ -212,6 +212,19 @@ open class SocketIOClient: NSObject, SocketIOClientSpec {
...
+    
+    /// Send an event to the server, with optional data items and optional write completion handler.
+    ///
+    /// If an error occurs trying to transform `items` into their socket representation, a `SocketClientEvent.error`
+    /// will be emitted. The structure of the error data is `[eventName, items, theError]`
+    ///
+    /// - parameter event: The event to send.
+    /// - parameter items: The items to send with this event. May be left out.
+    /// - parameter completion: Callback called on transport write completion.
+    open func emit(_ event: String, _ items: [SocketData], completion: (() -> ())?) {
```
nuclearace, line 226: Could you add `withItems` external param?

nuclearace (changes requested):
Thanks! I didn't think about this

Source/SocketIO/Client/SocketIOClient.swift
OneSman7: Wouldn’t it be more “swifty” to do `open func emit(_ event: String, with items: [SocketData], completion: (() -> ())?)` then?

nuclearace: Yeah that'd also be fine

OneSman7: Done!

nuclearace (approved):

PR #1338, opened 2021-03-25
Updated a example code

tsomaev:
I decided that it is better to show how to correctly implement the connection to sockets using the class, so that there are no questions on the similarity `Why isn’t my event handler being called?`

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,28 +7,35 @@ Socket.IO-client for iOS/OS X.
 ```swift
 import SocketIO
 
-let manager = SocketManager(socketURL: URL(string: "http://localhost:8080")!, config: [.log(true), .compress])
-let socket = manager.defaultSocket
-
-socket.on(clientEvent: .connect) {data, ack in
-    print("socket connected")
-}
-
-socket.on("currentAmount") {data, ack in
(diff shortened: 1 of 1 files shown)


tsomaev:
Also, I have a question. 

In our backend we use socket.io 4.0.0, but on iOS client used socket.io-client-swift version 16 and it correct work with sockets from backend, although it is said that the client supports socket.io 3+

tsomaev:
hi @nuclearace, will you merged this changes ?

README.md
```diff
@@ -7,28 +7,35 @@ Socket.IO-client for iOS/OS X.
 ```swift
 import SocketIO
 
-let manager = SocketManager(socketURL: URL(string: "http://localhost:8080")!, config: [.log(true), .compress])
-let socket = manager.defaultSocket
+class SocketIOManager{
+    
+    let manager = SocketManager(socketURL: URL(string: "http://localhost:8080")!, config: [.log(true), .compress])
```
CarolineLaw, line 12: it's better to safely unwrap stuff like URL's, you don't have to, but I'd suggest it :D

PR #1359, opened 2021-06-17
Fix swift warnings

WyattMufson:
In
```
case let .disconnected(reason, code):
```
both `reason` and `code` were not being used.

In
```
public protocol SocketEngineSpec: class {
```
you should be using `AnyObject` instead of `class`

In
```
public protocol SocketManagerSpec : AnyObject, SocketEngineClient {
```
`AnyObject` was redundant


WyattMufson:
@nuclearace this fixed a number of warnings

PR #1370, opened 2021-08-09
FIX/ObjC-6.0.1-Bridge

berinhardt:
FIX for issue-1358:

Exposes following API to Obj-C:
SocketManager.defaultSocket
SocketManager.disconnect

SocketIOClient.on(callback:)
SocketIOClient.connect()
SocketIOClient.emit(with:)


nuclearace:
Objc support was dropped

PR #1375, opened 2021-08-19
Handling reconnect scenarios properly when socket is hung

markshiz:
In situations where the socket connection is lost via a `SO_ERROR`, even though reconnect is enabled, the socket manager hangs the socket until a manual disconnect.  Symptom would be: the app would output something like the following:

```
2019-01-09 12:15:49.005630+0000 App [13746:3543026] [] nw_socket_handle_socket_event [C10.1:1] Socket SO_ERROR [61: Connection refused]
2019-01-09 12:15:49.016640+0000 App [13746:3543026] [] nw_socket_handle_socket_event [C10.2:1] Socket SO_ERROR [61: Connection refused]
```

Socket-IO would be logging the retry requests over and over again with the following log:
```
Tried connecting an already active socket
```

In this case, the `active` socket is hung, so a retry would never be attempted.  This condition can be simulated by running an app on a simulator while turning on Network Link Conditioner with 100% packet loss profile.  

This fix simply recycles the engine in the case that a `connecting` status has 1 or more retry attempts.  This seems to resolve the issue.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,5 @@ Socket.IO-Test-Server/node_modules/*
 .idea/
 docs/docsets/
 docs/undocumented.json
+
+.swiftpm
--- a/Source/SocketIO/Client/SocketIOClient.swift
+++ b/Source/SocketIO/Client/SocketIOClient.swift
@@ -150,7 +150,8 @@ open class SocketIOClient: NSObject, SocketIOClientSpec {
 
         manager.handleQueue.asyncAfter(deadline: DispatchTime.now() + timeoutAfter) {[weak self] in
             guard let this = self, this.status == .connecting || this.status == .notConnected else { return }
-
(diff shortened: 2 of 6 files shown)


WestFlow127:
@nuclearace this PR fixes the issue for [1415#](https://github.com/socketio/socket.io-client-swift/issues/1415)

This bug is critical. Can we please get this merged?

nuclearace:
@WestFlow127 Would you be able to test on development if I merge? I don't have capacity this week

WestFlow127:
@nuclearace Yes, I can do that.

WestFlow127:
Thank you! @nuclearace

ricardopereira:
@WestFlow127 did it passed your testing?
Also, any idea when this fix will be released? Thanks

WestFlow127:
@ricardopereira Yes, I have just recently fully tested this against my companies socket server.
My testing passed, and worked well.

ricardopereira:
I also tested it. I was facing issues with reconnections and this PR fixes it. Thank you!

PR #1385, opened 2021-10-30
Added support for Mac Catalyst

NachoSoto:


NachoSoto:
Note that this currently doesn't work yet because the project depends on `Starscream`'s `.framework`, which isn't compatible with MacCatalyst. The project could be updated to use an explicit dependency on the code, or changed to use XCFrameworks instead. Support for that is coming to Carthage soon.

nuclearace:
Can you open against development?

NachoSoto:
Sure, though that branch seems to be 6 commits behind `master`?

PR #1436, opened 2023-02-16
Fix documentation error

Ricardo-all-win-software:
In the documentation apears repeated the word "when" 2 times in 3 parts

image


nuclearace:
To properly fix, you'd want to fix place in code and regen docs

PR #1444, opened 2023-04-25
Fix the network switch cases by adding case for  .viabilityChanged(is…

aadilHaffez:
…Viable):


WestFlow127:
@aadilHaffez Can you explain what this does and provide an example?
I'd like to further understand this.

PR #1451, opened 2023-08-21
Fix Starscream to version 4.0.4

IvanGamov:


tumata:
Could someone instead of fixing the version to `4.0.4` fix `socket.io-client-swift` so that it stops supporting iOS 10?

This is causing our pod to fail during `pod push` because we cannot use the `podfile`to override the minimum deployment version of the socket.io target (we're building a pod, not a top-level app).

Anyways iOS 10 isn't supported by Xcode anymore.

tumata:
Could someone instead of fixing the version to `4.0.4` fix `socket.io-client-swift` so that it stops supporting iOS 10?

This is causing our pod to fail during `pod push` because we cannot use the `podfile`to override the minimum deployment version of the socket.io target (we're building a pod, not a top-level app).

Anyways iOS 10 isn't supported by Xcode anymore.

@IvanGamov

PR #1452, opened 2023-08-22
fix for cocoa pod version

romaHerman:
locked Starscream to the 4.0.4


nuclearace:
Can you also change the ios deployment target? Not sure what it should be set to now

romaHerman:
not sure we should change anything but starscream 
because it's the only thing that was causing an issue

nuclearace:
@romaHerman See https://github.com/socketio/socket.io-client-swift/issues/1453. Would be nice if you could bump to 11.0 to reflect starscream

romaHerman:
bumped iOS and tvOS versions

nuclearace:
@romaHerman I appreciate the changeset, but I went ahead and updated the library to support 4.0.6. Can you see if 16.1.0 release version works for you? (Be sure to clear any previous build data)

PR #1465, opened 2023-11-17
Update Socket.IO-Client-Swift.podspec

fdcfujihara:


fdcfujihara:
give different pod module name

PR #1473, opened 2024-02-02
Improved websocket reconnection logic

jonstaff:
This PR contains two primary improvements:

First, a configurable timeout was added for when the `SocketEngine`
initially attempts to establish a connection via websocket. Without this
timeout in place, the `SocketEngineClient` (read: the `SocketManager`)
doesn't get properly informed of the hanging connection and wastes its
reconnect attempts. 10 seconds was chosen as the default timeout based
on extensive testing with poor network conditions, but can be modified
based on the requirements of the app using this library if necessary.

Second, a configurable heartbeat mechanism was added to the
`SocketEngine` implementation. This allows consumers of this library to
specify a socket endpoint to use for heartbeat requests. Since
Socket.IO's ping/pong implementation was reversed in v3+ of the
protocol, clients no longer have a liveness check built in. The ping and
pong events are reserved for Socket.IO, so this allows you to specify a
different (hopefully very simple and very performant) event that can be
used for the same purpose.

Instead of mirroring the ping/pong logic exactly, the heartbeat is
designed to only beat when other packets aren't being transmitted.
Meaning, if your socket is already actively receiving data, there's no
reason to perform a heartbeat, since it's obviously connected. This is
done to reduce the load on the server.


jonstaff:
Closing this PR since it doesn't seem likely to get merged and I've started to add additional changes into my fork that are out of the scope of this PR (and somewhat specific to my project). Hopefully this is useful to others who've encountered similar issues.

PR #1486, opened 2024-03-21
Update Starscream to v4.0.8

gin0606:
ref https://github.com/socketio/socket.io-client-swift/issues/1478

We need to update the version of Starscream to support the Privacy Manifest.

https://github.com/daltoniam/Starscream/compare/4.0.6...4.0.8

Issue #1478, opened 2024-03-15
Starscream version (4.0.8)


doubleJN:
Hi!

I am using SPM to utilize SocketIO.
To use Starscream with Privacy Manifest applied, version 4.0.8 is required. However, in the latest SocketIO (16.1.0), there is a dependency on version 4.0.6, so it is impossible for me to use Starscream version 4.0.8.

youngToMaturity:
Same issue. need update.

benjamin-es-hall:
Any update on this?

cbiggin:
There is an outstanding PR #1487 for this.

PR #1464, opened 2023-11-13
v16.1.1

nuclearace:
Fixes for various connection status stuff

PR #1487, opened 2024-03-22
Update Starscream to v4.0.8

gin0606:
ref https://github.com/socketio/socket.io-client-swift/issues/1478

We need to update the version of Starscream to support the Privacy Manifest.

https://github.com/daltoniam/Starscream/compare/4.0.6...4.0.8

This pull request is a redo of https://github.com/socketio/socket.io-client-swift/pull/1486. 

I have reconfirmed the operation, and it appears to be functioning correctly in my application( socket.io 2 servers ).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -8,7 +8,7 @@ let package = Package(
         .library(name: "SocketIO", targets: ["SocketIO"])
     ],
     dependencies: [
-        .package(url: "https://github.com/daltoniam/Starscream", .exactItem("4.0.6")),
+        .package(url: "https://github.com/daltoniam/Starscream", .exactItem("4.0.8")),
     ],
     targets: [
         .target(name: "SocketIO", dependencies: ["Starscream"]),
--- a/Cartfile
+++ b/Cartfile
@@ -1,1 +1,1 @@
-github "daltoniam/Starscream" ~> 4.0.6
+github "daltoniam/Starscream" ~> 4.0.8
--- a/Cartfile.resolved
+++ b/Cartfile.resolved
@@ -1,1 +1,1 @@
-github "daltoniam/Starscream" "4.0.6"
+github "daltoniam/Starscream" "4.0.8"
--- a/Package.resolved
+++ b/Package.resolved
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/daltoniam/Starscream",
         "state": {
           "branch": null,
-          "revision": "ac6c0fc9da221873e01bd1a0d4818498a71eef33",
-          "version": "4.0.6"
+          "revision": "c6bfd1af48efcc9a9ad203665db12375ba6b145a",
+          "version": "4.0.8"
         }
       }
     ]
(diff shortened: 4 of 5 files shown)


slice-varundudeja:
when will this get released? its needed to remove apple warnings.

bmt2018:
With the privacy manifest deadline looming, if there isn't a new release that merges these changes for the Starscream dependency and you're using SPM, it looks like it might(?) be possible to fix this by bringing in the Socket.IO client as a local package and then editing the package.swift file to specify Starscream 4.0.8 instead of 4.0.6.  I'm basing this assumption on https://developer.apple.com/documentation/xcode/editing-a-package-dependency-as-a-local-package.  Has anyone tried this?

amuyu:
`Starscream` needs to be updated because it is included in the following content.
I'm curious about when it will be reflected.  (cc. @nuclearace )
  
> The following are commonly used SDKs in apps on the App Store. Starting in spring 2024, you must include the privacy manifest for any SDK listed below when you submit new apps in App Store Connect that include those SDKs
https://developer.apple.com/support/third-party-SDK-requirements/  
  
   https://github.com/daltoniam/Starscream/pull/1008

cbiggin:
This really does need to be updated because you've specified `Exact Version` (of Starscream v4.0.6) instead of `Up to Next Major Version` in the `Package.swift`. This means that I cannot even explicitly add Starscream v4.0.8 as an SPM package since it conflicts.

WestFlow127:
@nuclearace this should be added to [v16.1.1](https://github.com/socketio/socket.io-client-swift/pull/1464)

imnak0-naver:
is there anyone who can approve this PR? 'o'?

Package.swift
```diff
@@ -8,7 +8,7 @@ let package = Package(
         .library(name: "SocketIO", targets: ["SocketIO"])
     ],
     dependencies: [
-        .package(url: "https://github.com/daltoniam/Starscream", .exactItem("4.0.6")),
+        .package(url: "https://github.com/daltoniam/Starscream", .exactItem("4.0.8")),
```
cbiggin, line 11: I think this should be `upToNextMajor`.

gin0606: I also believe that's the way to go. However, since I don't fully understand the intentions behind the changes made in https://github.com/socketio/socket.io-client-swift/pull/1451/commits/3c81d0ee55111a80a8f39b9c61782887ec98ceb4 , I feel it's not appropriate for me to do it.

EgzonArifi:
any updates on this?

benjamin-es-hall:
The deadline requiring privacy manifests has now past and expecting releases including socket-io to get blocked by apple unless this is done. Any update on when this will be merged and deployed?

tumata:
@gin0606 Please note that updating `Starscream` while not also updating `s.ios.deployment_target = '12.0'` leads to crashes.

See #1488 - just merge that into this PR. This is completely blocking our team.

gin0606:
The changes in https://github.com/socketio/socket.io-client-swift/pull/1492 are better, so this Pull Request is no longer necessary.

PR #1491, opened 2024-05-09
Fix the issue where it takes over 60 seconds to close the socket after a network disconnection

hsjcom:
Fix the issue where it takes over 60 seconds to close the socket after a network disconnection, and the problem where the server-side socket takes over 30 seconds to close when the server ends a connection


davidkessler-ch:
I think this PR should definitely be merged...

PR #1492, opened 2024-05-14
Support Privacy Manifest: bump Starscream package version to upToNextMajor 4.0.8

xxZap:
This PR solves the impossibility of **publishing** iOS applications due to the [Privacy Manifest](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files) rule which is now mandatory.

The problem is that **socket.io** is pointing to **4.0.6** **Starscream**, but Starscream started supporting the Privacy Manifest policy from **4.0.8**.

This PR updates its minimum version from 4.0.6 to `4.0.8`.

I already published my application to the App Store with this change without any issues.

⭐️ Eventually, instead of using `upToNextMajor` you can use `exact`. It's up to you.

### DIFF
--- a/.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
+++ b/.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
+   version = "1.0">
+   <FileRef
+      location = "self:">
+   </FileRef>
+</Workspace>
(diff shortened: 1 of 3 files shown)


9oHigh:
It's been a month, and it's still not merged? 
sad : (

9oHigh:
@nuclearace Please accept PR..

icepat:
Would be amazing if you guys could merge this PR
We have Starscream crashing in the background with WalletConnect and are hoping that the 4.0.8 will solve this.

xxZap:
Any news about this PR?
I think that this easily solves a problem that affects everyone who is using socket.io

PR #1494, opened 2024-06-11
fix: 升级依赖

yuanyeer:


yuanyeer (commented):
yes

cbiggin:
Out of curiousity I looked at this PR and you are proposing changing the repo to your own private repository. I don't think so!!

Not sure if you did this deliberately or accidentally but in any case, this PR should be discarded/deleted.

nuclearace:
I don't think we want to change the homepage

PR #1526, opened 2026-04-24
Phase 1: SocketIOClientOption.autoConnect

howardsun-tw:
## Summary
- Adds `SocketIOClientOption.autoConnect(Bool)` (default `false` — JS inverts to `true`).
- When `true`, `SocketManager.init` calls `defaultSocket.connect()` then opens the engine, so the default namespace is auto-joined when the engine open completes (mirrors JS `io(url, {autoConnect: true})`).
- Only `defaultSocket` is auto-CONNECTed; non-default namespaces still require explicit `socket.connect()` (matches JS — `Manager.autoConnect` only opens the engine, not arbitrary namespaces).
- Wired through both init APIs: array literal config and the `@objc` `[String: Any]?` config dictionary.

## Spec
docs/superpowers/specs/2026-04-24-socketio-gap-fill-design.md — Phase 1.

## Test plan
- [x] Unit: option round-trip + description + value extraction
- [x] Unit: dict-init parser recognizes `"autoConnect"` (true/false)
- [x] Unit: `autoConnect` property defaults `false`; `setConfigs` parses
- [x] Unit: `init` triggers `connect()` only when `autoConnect == true`; engine constructed when true, not when false
- [x] E2E: `defaultSocket` auto-joins
- [x] E2E: `/admin` namespace does NOT auto-join until explicit `connect()`
- [x] E2E: `autoConnect:false` leaves `defaultSocket` in `.notConnected`
- [x] Full suite: 171/171

## JS-divergence policy
None. Implementation matches JS reference (`socket.io-client/lib/manager.ts` constructor calls `this.open()` synchronously when `autoConnect: true`). Default flipped from `true` to `false` to preserve existing Swift caller behavior — documented as the only divergence.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


howardsun-tw:
Closing — opened in error. Local development only.

